PR #41, opened 2026-03-04
feat: VSCode extension Phase 0+1 — compiler API, syntax highlighting, diagnostics

thiagoralves:
## Summary

- **Phase 0 — Compiler API expansion**: Export 60+ types from `src/index.ts`, add `analyze()` function that returns partial results even with errors (for LSP), add `endLine`/`endColumn` to `CompileError`, create `src/ast-utils.ts` with AST traversal utilities (`walkAST`, `findNodeAtPosition`, `findInnermostExpression`, `collectReferences`)
- **Phase 1 — Extension scaffold + language server**: TextMate grammar for full ST syntax highlighting, language client/server with IPC transport, document manager with workspace-wide `.st` file discovery for multi-file projects, automatic standard library loading (`iec-standard-fb.stlib`, `oscat-basic.stlib`), debounced diagnostics, test file detection to exclude `TEST`/`SETUP` files from source parsing
- All 1414 existing compiler tests pass with zero regressions

## Test plan

- [x] Compiler builds cleanly (`npm run build`)
- [x] Extension builds cleanly (`tsc -b` in `vscode-extension/`)
- [x] All 1414 compiler tests pass
- [x] Manual testing: F5 launch in VSCode shows syntax highlighting and real-time diagnostics
- [x] Manual testing: multi-file project (smart-traffic-light) resolves cross-file types and FBs
- [x] Manual testing: standard FB library (TON, CTU, etc.) recognized without errors
- [x] Manual testing: test files excluded from source parsing (no false `TEST` token errors)

### DIFF
--- a/src/ast-utils.ts
+++ b/src/ast-utils.ts
@@ -86,10 +86,11 @@ export function findNodeAtPosition(
 ): ASTNode | undefined {
   let best: ASTNode | undefined;
 
-  walkAST(ast, (node) => {
-    if (containsPosition(node, file, line, column)) {
-      best = node;
+  walkAST(ast, (node): boolean | void => {
+    if (!containsPosition(node, file, line, column)) {
+      return false; // prune subtrees that can't contain the position
     }
+    best = node;
   });
 
   return best;
@@ -106,8 +107,11 @@ export function findInnermostExpression(
 ): Expression | undefined {
   let best: Expression | undefined;
 
-  walkAST(ast, (node) => {
-    if (isExpression(node) && containsPosition(node, file, line, column)) {
+  walkAST(ast, (node): boolean | void => {
+    if (!containsPosition(node, file, line, column)) {
+      return false; // prune subtrees that can't contain the position
+    }
+    if (isExpression(node)) {
       best = node as Expression;
     }
   });
--- a/vscode-extension/server/src/diagnostics.ts
+++ b/vscode-extension/server/src/diagnostics.ts
@@ -20,7 +20,7 @@ export function toLspDiagnostic(error: CompileError): Diagnostic {
   const startLine = Math.max(0, (error.line ?? 1) - 1);
   const startCol = Math.max(0, (error.column ?? 1) - 1);
   const endLine = error.endLine != null ? error.endLine - 1 : startLine;
-  const endCol = error.endColumn != null ? error.endColumn - 1 : startCol;
+  const endCol = error.endColumn != null ? error.endColumn - 1 : startCol + 1;
 
   return {
     range: Range.create(
--- a/vscode-extension/server/src/document-manager.ts
+++ b/vscode-extension/server/src/document-manager.ts
@@ -240,6 +240,7 @@ function uriToFilePath(uri: string): string {
 function isTestFile(source: string): boolean {
   // Strip leading comments and whitespace, then check the first keyword
   const stripped = source
+    .replace(/^\uFEFF/, "")             // strip UTF-8 BOM
     .replace(/\/\/.*$/gm, "")           // remove line comments
     .replace(/\(\*[\s\S]*?\*\)/g, "")   // remove block comments
     .trimStart();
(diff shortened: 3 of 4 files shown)


vscode-extension/server/src/diagnostics.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Convert a CompileError to an LSP Diagnostic.
+ * STruC++ uses 1-indexed line/column; LSP uses 0-indexed.
+ */
+export function toLspDiagnostic(error: CompileError): Diagnostic {
+  const startLine = Math.max(0, (error.line ?? 1) - 1);
+  const startCol = Math.max(0, (error.column ?? 1) - 1);
+  const endLine = error.endLine != null ? error.endLine - 1 : startLine;
+  const endCol = error.endColumn != null ? error.endColumn - 1 : startCol;
```
thiagoralves: **Zero-width diagnostic ranges produce invisible squiggles**

When \`endLine\`/\`endColumn\` are not set (common for many compiler errors), both start and end resolve to the same position, creating a zero-width range. Most editors won't show any visible underline for this.

Minimal fix — extend by at least one character:
```ts
const endCol = error.endColumn != null ? error.endColumn - 1 : startCol + 1;
```

Or for better UX, highlight to end-of-line when the end position is unknown (use \`Number.MAX_SAFE_INTEGER\` for the end column — the LSP client will clamp it).

src/ast-utils.ts
```diff
@@ -0,0 +1,586 @@
...
+ * Optionally filter by scope (e.g., "MyProgram").
+ */
+export function collectReferences(
+  ast: CompilationUnit,
+  symbolName: string,
+  scope?: string,
+): ASTNode[] {
+  const refs: ASTNode[] = [];
+  const upperName = symbolName.toUpperCase();
+  let currentScope: string | undefined;
```
thiagoralves: **\`collectReferences\` scope tracking never resets between sibling POUs**

\`currentScope\` is set when entering a ProgramDeclaration/FunctionDeclaration/etc. but never restored when leaving that scope. Since \`walkAST\` visits siblings sequentially, after visiting \`PROGRAM A\`, \`currentScope\` remains \`\"A\"\` until the next POU overwrites it.

This causes two problems:
1. In a file with \`PROGRAM A\` followed by \`PROGRAM B\`, if you search with \`scope=\"B\"\`, references in B's var blocks (visited before B's scope-setting node) will be filtered out since \`currentScope\` is still \`\"A\"\`.
2. For methods inside FBs, \`currentScope\` gets set to the method name and never reverts to the FB name. Searching with \`scope=\"MyFB\"\` will miss references inside methods.

Fix: save/restore scope around the walk, or use a scope stack:
```ts
const previousScope = currentScope;
currentScope = (node as ...).name;
// ... (let walkAST recurse into children)
// After children are visited, restore:
// But walkAST doesn't give a post-visit hook...
```

Since \`walkAST\` only has a pre-visit callback, the cleanest fix is to either add a post-visit callback to \`walkAST\`, or restructure \`collectReferences\` to manually recurse instead of using \`walkAST\`.

thiagoralves (commented):
## Code Review — Round 2

Good work addressing the first round of feedback. The `runPipeline` extraction, type-safe AST casts, `vscode-uri` adoption, discovery caching, and `analyze()` tests all look solid.

### Issues found (6 inline comments posted)

| Priority | File | Issue |
|----------|------|-------|
| **High** | `server.ts:73` | `didChangeWorkspaceFolders` params shape doesn't match LSP spec — `params.added`/`params.removed` should be `params.event.added`/`params.event.removed`. Workspace folder changes are silently ignored. |
| **Medium** | `diagnostics.ts:23` | Zero-width diagnostic ranges when `endLine`/`endColumn` are unset produce invisible squiggles. |
| **Medium** | `ast-utils.ts:129` | `collectReferences` scope tracking never resets — causes incorrect filtering in multi-POU files and methods inside FBs. |
| **Low** | `ast-utils.ts:89` | `findNodeAtPosition` walks entire AST without pruning non-containing subtrees. |
| **Low** | `index.ts:509` | `analyze()` returns a fresh `StdFunctionRegistry` disconnected from the one used during analysis. |
| **Low** | `document-manager.ts:245` | `isTestFile` doesn't handle UTF-8 BOM. |

### Recommendation

Fix the **High** and **Medium** issues before merge. The low-priority items are fine to defer to Phase 2.

PR #76, opened 2026-03-06
fix: missing beforeEach vitest imports in 3 test files (#70)

michaeldistel:
Closes #70

## Summary

- Add `beforeEach` to vitest import in `tests/semantic/std-function-registry.test.ts`, `tests/semantic/type-checker.test.ts`, and `tests/backend/codegen.test.ts`
- Fixes `ReferenceError` at runtime caused by `beforeEach` being used without being imported

## Note on extra diff noise

The lint-staged hook runs Prettier automatically on every staged file. Touching any of these files caused Prettier to reformat them: single quotes to double quotes and long lines wrapped to fit print width. These are cosmetic changes only, no logic was altered. The sole intentional change per file is the addition of `beforeEach` to the vitest import line.

## Testing

- `npm test`: 51 test files passed, 1414 tests passed, 7 skipped


thiagoralves:
Closing as obsolete — this fix isn't needed.

The PR addresses a claimed runtime `ReferenceError` from `beforeEach` being used without an import in three test files. But `vitest.config.ts` sets `globals: true`, which injects `beforeEach` / `describe` / `it` / `expect` as globals — so the missing import was never actually a runtime error under this config.

Verified against current `development` (without this PR applied): all three files pass — `tests/semantic/std-function-registry.test.ts`, `tests/semantic/type-checker.test.ts`, `tests/backend/codegen.test.ts` → 45/45 tests pass, no `ReferenceError`.

The only functional change here is adding an unnecessary import; the rest of the diff is Prettier churn (as noted in the PR description). Not worth merging. Closing #70 as well.

PR #77, opened 2026-03-06
fix: replace dist/ with src/ in test imports (#71)

michaeldistel:
Closes #71

## Summary

- 8 test files imported from `dist/` causing tests to silently validate stale compiled output instead of current source
- Replace `dist/` with `src/` in all import paths; Vitest's Vite pipeline transpiles `.ts` on-the-fly so `dist/` is never needed
- `cli-library.test.ts` left unchanged — it invokes `dist/cli.js` as a subprocess which correctly requires a built binary

## Testing

`npm test`: 51 test files passed, 1414 tests passed, 7 skipped — no regressions

### DIFF
--- a/tests/backend/codegen-composite.test.ts
+++ b/tests/backend/codegen-composite.test.ts
@@ -12,9 +12,14 @@
  */
 
 import { describe, it, expect } from "vitest";
-import { compile } from "../../dist/index.js";
+import { compile } from "../../src/index.js";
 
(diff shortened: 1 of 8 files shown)


thiagoralves:
Superseded by #153. #153 converts all 14 current dist-importing test files (this PR covered 8 of the then-9 and conflicts against current development), and adds a guard test + a deliberate package-smoke test. The real justification is coverage attribution: dist/ imports contribute 0% to src/ coverage (dist is excluded and not remapped), not the 'fails without build' reason here (globalSetup runs tsc first).

PR #81, opened 2026-03-06
refactor: remove internal duplicate test sections in codegen files (#73)

michaeldistel:
## Summary
- Remove 5 duplicate \`it\` blocks from "Validation Examples" section in \`codegen-control-flow.test.ts\` (nested IF/ELSIF, FOR+EXIT, WHILE, REPEAT-UNTIL, CASE with ranges)
- Remove entire "Phase 3.3 Docs" \`describe\` block from \`codegen-composite.test.ts\` (array access, struct access, array-of-struct)

Closes #73

## Testing
- \`npm test\`: 51 files, 1406 passed, 7 skipped — all passing
- \`npm run build\`: clean

### DIFF
--- a/tests/backend/codegen-composite.test.ts
+++ b/tests/backend/codegen-composite.test.ts
@@ -14,7 +14,12 @@
 import { describe, it, expect } from "vitest";
 import { compile } from "../../dist/index.js";
 
-function compileST(source: string): { cppCode: string; headerCode: string; success: boolean; errors: unknown[] } {
+function compileST(source: string): {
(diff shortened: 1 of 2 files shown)


thiagoralves:
Superseded by #152. #152 removes the same 8 duplicate 'validation example' tests, but freshly against current development (this PR conflicts), without re-adding the now-present compileST helper, and it preserves the one non-redundant assertion (CASE numeric range expanding to intermediate labels) that a blind delete here would have dropped.

PR #82, opened 2026-03-06
move references.test.ts to frontend/ (#74)

michaeldistel:
## Summary

- Move `tests/semantic/references.test.ts` → `tests/frontend/references.test.ts`; file only calls `parse()`, no semantic analysis

## Testing

- All 51 test files pass (1414 passed, 7 skipped) after the move
- `tests/frontend/ast-utils.test.ts` does not exist on `development`; that part of the issue is moot on this branch

Closes #74


thiagoralves:
Superseded by #150. This PR is correct but incomplete — it moves only references.test.ts (semantic → frontend), while #74 also calls for moving ast-utils.test.ts (frontend → integration, since it runs the full analyze() pipeline). #150 does both moves as pure git renames and fully closes #74. Also, this branch is based on a March commit.

PR #83, opened 2026-03-06
fix: fragile/vacuous assertions in oscat, codesys-import, located-variables tests (#75)

michaeldistel:
Closes #75

## Summary

- Replace silent `return` with `expect.fail()` in oscat second test so it hard-fails if transpilation output is missing
- Replace brittle exact-equality FUNCTION/TYPE count assertions in codesys-import V3 vs V2.3 comparison with `toBeGreaterThan(0)`
- Add `LOCATED_VAR_IN_FB` error code to semantic analyzer; thread `code` through `addError` and `index.ts` error mapping; assert on `e.code` in located-variables test instead of fragile string-contains

## Testing

- `npm test`: 51 test files, 1414 passed, 7 skipped — all passing

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -297,6 +297,7 @@ export function compile(
       line: err.line ?? 0,
       column: err.column ?? 0,
       severity: "error",
+      ...(err.code !== undefined ? { code: err.code } : {}),
     });
   }
   for (const warn of semanticResult.warnings) {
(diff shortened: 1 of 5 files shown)


thiagoralves:
Superseded by #149. Re-checking #75 against current `development`: item 2 (codesys-import V3-vs-V2.3 exact equality) is already fixed — it uses `toBeGreaterThan(...)` tolerance checks now. #149 fixes the two still-valid items (oscat vacuous `return` → `expect.fail`; located-variables generic `'FUNCTION_BLOCK'` match → anchored to the distinctive rule text). This PR is stale and conflicts against current development (it also touched src/index.ts and analyzer.ts, and the already-fixed codesys test).

PR #202, opened 2026-08-04
feat(codegen): emit locatedGlobals[] and tighten located-address validation

thiagoralves:
## Why

`locatedVars[]` mixes two ownership classes: POU-local `VAR ... AT`, serviced by the owning IEC task around `run()`, and `CONFIGURATION VAR_GLOBAL ... AT`, owned by no task and therefore serviced by a host dispatcher at a quiescent frame boundary. The descriptors carry no scope, and a runtime cannot recover it — globals are file-scope `GlobalVar<V>` singletons that appear nowhere in the object graph reachable through `ConfigurationInstance`.

So the OpenPLC v4 runtime inferred the split from **array position**, assuming `[program-local …][config globals …]`. We emit config globals **first** (`emitFileScopeGlobals` runs before the program headers), so the program-local block is always the tail: one POU-local located variable made the runtime's count collapse to zero and silently stop servicing **every** located global. Modbus writes reached the image but never the program; `%IX`/`%QX`/`%MW` globals were equally affected.

Reported on the forum as *"%MX locations now invalid - Runtime v4"* — the reporter's `%MX` writes worked until he added one unrelated located variable to a POU, and downgrading to runtime v4.1.3 (which predates the threaded process-image model) fixed it.

## What this does

**1. Emit `locatedGlobals[]`** — the canonical storage pointer of every located `VAR_GLOBAL`, the same `raw_ptr()` value written into `locatedVars[].pointer`. A runtime resolves the config-scope entries by **pointer identity**, with no dependence on array order. Located variables are always elementary types, so every entry is a scalar `IECVar` and `raw_ptr()` exists. Populated in the configuration constructor because `raw_ptr()` is not a constant expression, and to keep it beside the `locatedVars[]` population it must agree with.

The `extern "C"` accessors are emitted **here rather than left to a host runtime's shim**. That is what makes this backwards-compatible in both directions: the OpenPLC runtime rebuilds its shim against each upload's generated header, so a shim referencing these symbols would fail to *compile* against an older project, whereas an accessor emitted beside its own array always matches. An older runtime never resolves the symbols and is bit-for-bit unaffected; a newer runtime probes with `dlsym` and skips located globals when they are absent.

Guarded by `STRUCPP_THREADED` throughout — freestanding targets bind every located variable directly and have no dispatcher, so they pay nothing. Verified that a non-threaded build emits no `locatedGlobals` symbol at all.

**2. Close three gaps in `validateLocatedVariables()`**, each of which previously produced silent misbehaviour rather than a diagnostic:

- **Configuration `VAR_GLOBAL`s were invisible to every located rule.** They live in `ast.configurations[].varBlocks` while `buildVarBlockSymbols` only walks POU scopes, so a POU-local `VAR ... AT %MX0.0` alongside a `VAR_GLOBAL ... AT %MX0.0` compiled — and the two entries resolve to one image slot serviced by two different paths. Now collected (deduped by name, since codegen emits one file-scope singleton per name and a global repeated across configurations is one canonical global).
- **A located variable in a PROGRAM instantiated more than once** was accepted. Codegen allocates one slot per *declaration* and the pointer is assigned `this`-relative in the POU constructor — the class cannot name its own instance — so the last instance constructed wins and the others are never serviced. Same reasoning as the existing `FUNCTION_BLOCK` restriction. IEC 61131-3 permits multiple program instances and answers per-instance addressing with a partly specified location (`AT %I*`) resolved by `VAR_CONFIG`, which is unsupported here, so the fully specified form must be rejected.
- **Only `VAR` and `VAR_GLOBAL` may now be located.** Interface sections describe a call contract, not hardware, and a located interface declaration hands the runtime two owners for one image slot — a hazard invisible from the generated code. `VAR_EXTERNAL` is the sharpest case: codegen emits it as `GlobalVar<T>*` and collects located vars from local declarations only, so the address is silently dropped while duplicating the global's claim. Matches the editor (`DISALLOWED_LOCATION_CLASSES`, GH issue #904).

`addressKey()` is deliberately left matching on exact `(area, size, byte, bit)` and this is now documented on the function. The image is **not** flat memory: each size class has its own array in the runtime (`bool_memory[][]`, `int_memory[]`, `dint_memory[]`, `lint_memory[]`) and `byte_index` indexes that array, so `%MW0` and `%MD0` name unrelated storage rather than overlapping bytes.

## Behaviour change — read before merging

The validation rejects projects that previously compiled. Editor-authored projects are unaffected (the editor already blocks located interface classes), but hand-written ST may be. `tests/fixtures/smart-traffic-light` declared its located I/O in `VAR_INPUT`/`VAR_OUTPUT` and was moved to `VAR`, which is where it belonged. Each new error surfaces a program that previously looked fine and did not work.

## Verification

- **2008 tests pass** (79 files), 15 new.
- Emission verified against a real editor-exported project; a non-threaded build emits no `locatedGlobals`.
- End-to-end on hardware (Raspberry Pi, runtime built from the companion PR): a Modbus coil write to `%MX0.0` now reaches program logic and the located `%MD0` global is committed back — `2 config-scope shared globals` where the old code reported `0`.
- Backwards compatibility exercised both ways: a bundle built from pre-change STruC++ makes the new runtime warn and run with globals inert; the program still runs and POU-local located I/O is unaffected.

## Release coupling

The editors pin **v0.6.2**, so the release must be cut as `v0.6.2`.

---

### Related PRs — merge together

This fix spans four repos. All target `development`.

| Repo | PR | Contents |
|---|---|---|
| STruCpp | Autonomy-Logic/STruCpp#202 | emits `locatedGlobals[]` + accessors; tightens located-address validation |
| openplc-runtime | Autonomy-Logic/openplc-runtime#161 | consumes it via pointer-identity join |
| openplc-editor | Autonomy-Logic/openplc-editor#983 | pins STruC++ v0.6.2 |
| openplc-web | Autonomy-Logic/openplc-web#643 | pins STruC++ v0.6.2 (shared parity) |

**Ordering:** the runtime degrades gracefully against an older STruC++ (warns, skips located globals, program still runs), and an older runtime ignores the new symbols entirely — so merge order does not matter. But the fix only becomes active for a project **re-exported** from an editor shipping v0.6.2, so all four need to land before it reaches users.

### DIFF
--- a/src/backend/codegen.ts
+++ b/src/backend/codegen.ts
@@ -5868,6 +5868,19 @@ export class CodeGenerator {
       `constexpr uint32_t locatedVarsCount = ${this.locatedVars.length};`,
     );
     this.emitHeader("");
+
+    // Located CONFIGURATION VAR_GLOBALs — see generateLocatedGlobalsDefinition()
+    // for why the scope has to be stated rather than inferred by the runtime.
+    const globalCount = this.locatedVars.filter(
+      (v) => v.programName === "@config",
+    ).length;
+    this.emitHeader("#ifdef STRUCPP_THREADED");
+    this.emitHeader(
+      `extern void *locatedGlobals[${globalCount === 0 ? 1 : globalCount}];`,
+    );
+    this.emitHeader(`constexpr uint32_t locatedGlobalsCount = ${globalCount};`);
+    this.emitHeader("#endif");
+    this.emitHeader("");
   }
 
   /**
@@ -5908,6 +5921,78 @@ export class CodeGenerator {
     }
 
     this.emit("};");
+    this.emit("");
+
+    this.generateLocatedGlobalsDefinition();
+  }
+
+  /**
+   * Emit the located-globals pointer array plus its C-linkage accessors.
+   *
+   * `locatedVars[]` mixes two ownership classes: POU-local `VAR ... AT` (serviced
+   * by the owning IEC task around run()) and CONFIGURATION `VAR_GLOBAL ... AT`
+   * (owned by no task, so a host dispatcher must copy them at a quiescent frame
+   * boundary). The descriptors carry no scope, and a runtime cannot recover it:
+   * globals are file-scope singletons that appear nowhere in the configuration
+   * object graph. Runtimes previously *inferred* the split from each entry's
+   * position in locatedVars[], which broke as soon as a POU declared a located
+   * variable and silently stopped servicing every located global.
+   *
+   * So state it here instead. locatedGlobals[] holds the canonical storage
+   * pointer of every located VAR_GLOBAL — the same raw_ptr() value written into
+   * locatedVars[].pointer — letting a runtime identify the config-scope entries by
+   * pointer identity, independent of array order. Located variables are always
+   * elementary types, so every entry is a scalar IECVar and raw_ptr() exists.
+   *
+   * Emitted only under STRUCPP_THREADED: freestanding targets bind every located
+   * variable directly and have no dispatcher, so they would pay RAM for nothing.
+   *
+   * The accessors are emitted here rather than left to a host runtime's shim so
+   * the generated code stays self-contained. A shim referencing these symbols
+   * would fail to COMPILE against an older project's generated header, whereas an
+   * accessor emitted beside its own array always matches. A runtime predating
+   * this simply never resolves the symbols and skips located globals.
+   */
+  private generateLocatedGlobalsDefinition(): void {
+    const globals = this.locatedVars.filter((v) => v.programName === "@config");
+    // C++ forbids zero-length arrays at namespace scope; emit a placeholder and
+    // report count 0, exactly as locatedVars[] does. That lets a runtime tell
+    // "accessors absent" (older project — cannot service globals) apart from
+    // "present but count 0" (project genuinely has no located globals).
+    const arrayLen = globals.length === 0 ? 1 : globals.length;
+
+    this.emit("#ifdef STRUCPP_THREADED");
+    this.emit(
+      "// Canonical storage pointers of the located CONFIGURATION VAR_GLOBALs.",
+    );
+    this.emit(
+      "// Populated in the configuration constructor (like locatedVars[] above).",
+    );
+    this.emit(`void *locatedGlobals[${arrayLen}] = {`);
+    if (globals.length === 0) {
+      this.emit(`    nullptr  // placeholder; locatedGlobalsCount is 0`);
+    } else {
+      for (let i = 0; i < globals.length; i++) {
+        const g = globals[i]!;
+        const comma = i < globals.length - 1 ? "," : "";
+        this.emit(`    nullptr${comma}  // ${g.varName} AT ${g.address}`);
+      }
+    }
+    this.emit("};");
+    this.emit("");
+    this.emit(
+      "// C linkage: a host runtime is built once and loads many .so files, so it",
+    );
+    this.emit(
+      "// cannot reach namespaced C++ symbols by mangled name portably.",
+    );
+    this.emit(
+      'extern "C" void *const *strucpp_get_located_globals(void) { return locatedGlobals; }',
+    );
+    this.emit(
+      'extern "C" uint32_t strucpp_get_located_global_count(void) { return locatedGlobalsCount; }',
+    );
+    this.emit("#endif  // STRUCPP_THREADED");
     this.emit("");
   }
 
@@ -5962,6 +6047,24 @@ export class CodeGenerator {
           `${indent}locatedVars[${index}].pointer = ${locVar.varName}${memberAccess}.raw_ptr();`,
         );
       }
+    }
+
+    // Configuration VAR_GLOBALs additionally record their storage pointer in
+    // locatedGlobals[], which is what lets a host runtime tell config-scope
+    // entries from POU-local ones without guessing from array position. Emitted
+    // here (rather than as a static initializer) because raw_ptr() is not a
+    // constant expression, and to keep it beside the locatedVars[] population it
+    // must agree with.
+    if (programName === "@config") {
+      this.emit("#ifdef STRUCPP_THREADED");
+      this.emit(`${indent}// Initialize located-global pointers`);
+      for (let g = 0; g < progVars.length; g++) {
+        const locVar = progVars[g]!;
+        this.emit(
+          `${indent}locatedGlobals[${g}] = ${locVar.varName}${memberAccess}.raw_ptr();`,
+        );
+      }
+      this.emit("#endif");
     }
   }
 
--- a/src/semantic/analyzer.ts
+++ b/src/semantic/analyzer.ts
@@ -102,7 +102,31 @@ function getCompatibleTypes(size: "X" | "B" | "W" | "D" | "L"): string[] {
 }
 
 /**
+ * Variable-block kinds that may carry a physical location ("AT %...").
+ *
+ * IEC 61131-3 allows located declarations in VAR and VAR_GLOBAL only — interface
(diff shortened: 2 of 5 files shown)


src/semantic/analyzer.ts
```diff
@@ -730,18 +773,104 @@ export class SemanticAnalyzer {
...
+   * and a `VAR_GLOBAL ... AT %MX0.0` could both claim the same image slot — and
+   * they are serviced by different paths (the owning task vs. the dispatcher at
+   * the quiescent frame boundary), which makes the outcome nondeterministic.
+   *
+   * Globals sharing a name across configurations are one canonical global (codegen
+   * emits a single file-scope singleton, deduping by name), so dedupe here too —
+   * otherwise a project declaring the same global in two configurations would
+   * report a spurious duplicate-address error against itself.
+   */
+  private collectConfigurationLocatedVars(
```
dcoutinho1328, line 792: :warning: **Coverage gap (pre-existing, adjacent to this PR)** — `collectConfigurationLocatedVars()` only walks `ast.configurations[].varBlocks`. Top-level `VAR_GLOBAL` blocks declared outside any `CONFIGURATION` (`ast.globalVarBlocks`, the "GVL file" path — see `ast-builder.ts:401-405`) are not gathered here, and `buildVarBlockSymbols()` (which enforces the new `LOCATABLE_BLOCK_TYPES` restriction) is never called on them either. A located variable in a top-level GVL block would still bypass every rule this PR adds — the exact class of gap this PR is closing for `CONFIGURATION VAR_GLOBAL`. Given this PR is specifically about making located-variable scope-tracking exhaustive, worth a follow-up (or at least a doc/TODO note) rather than leaving it as a silent gap.

src/backend/codegen.ts
```diff
@@ -5909,6 +5922,78 @@ export class CodeGenerator {
...
+    this.emit("};");
+    this.emit("");
+    this.emit(
+      "// C linkage: a host runtime is built once and loads many .so files, so it",
+    );
+    this.emit(
+      "// cannot reach namespaced C++ symbols by mangled name portably.",
+    );
+    this.emit(
+      'extern "C" void *const *strucpp_get_located_globals(void) { return locatedGlobals; }',
```
dcoutinho1328, line 5990: :warning: **Cross-repo wire contract, not locally verifiable** — `strucpp_get_located_globals`/`strucpp_get_located_global_count` form a new C ABI consumed by `openplc-runtime#161` (per the PR description, joined by pointer identity). That repo isn't checked out locally, so I could not verify the consumer's expected shape against its authoritative source, per the wire-contract review step. The PR description states this was verified end-to-end on real hardware against the companion runtime PR, so this is flagged for awareness/traceability rather than as a blocking finding — worth double-checking the signatures still match if `openplc-runtime#161` has moved since that verification.

dcoutinho1328 (approved):
Reviewed the located-globals emission + validation-tightening changes. Verified on the actual PR branch (checked out in an isolated worktree, not just the diff): full suite (2008 tests, 79 files) passes, `npm run typecheck` is clean, and `npm run lint` reports 0 errors (only pre-existing warning-style findings that match this file's existing conventions).

Traced the index alignment between `generateLocatedGlobalsDefinition()`'s `globals` filter and `generateLocatedVarPointerInit("@config", ...)`'s `progVars` filter — both derive from `this.locatedVars.filter(v => v.programName === "@config")`, and `emitFileScopeGlobals()` (which populates `@config` entries) runs before both header and impl emission, so the two filtered arrays are guaranteed to agree in order/length. `resource.programInstances` / `instance.programType` field names used in the new `countProgramInstantiations()` match the AST (`ast.ts`), and Rule 1b correctly gates only on `scopeType === "program"` so it doesn't interact with the FUNCTION_BLOCK rule or the new `"configuration"` scope. The `cpp-compile.test.ts` suite (unchanged, but parametrized over `-DSTRUCPP_THREADED`) already exercises the new code path end-to-end through g++, including the `void**` → `void *const*` qualification conversion in the new accessor.

Two non-blocking observations left as inline comments below. No blocking issues found.

dcoutinho1328:
# PR #202 Review

## Summary
Emits a `locatedGlobals[]` pointer array + `extern "C"` accessors so a threaded runtime can identify `CONFIGURATION VAR_GLOBAL` located variables by pointer identity instead of inferring the split from array position (which silently broke as soon as a POU declared its own located variable). Also closes three validation gaps: only `VAR`/`VAR_GLOBAL` may carry `AT` locations, a located variable can't live in a multiply-instantiated `PROGRAM`, and configuration-scope globals now participate in duplicate-address/type-compatibility checks alongside POU-local located variables.

## Module: Compiler (single-package repo)

### Checklist Results (default criteria — no repo checklist file present)
- :white_check_mark: **Correctness** — traced index alignment between the header's `locatedGlobals[N]` declaration, its definition, and the pointer-population loop; all three derive from the same `this.locatedVars.filter(v => v.programName === "@config")`, populated once by `emitFileScopeGlobals()` before any of the three run, so ordering is guaranteed to agree.
- :white_check_mark: **New validation logic** — `Rule 1b` (multiply-instantiated PROGRAM) correctly keys off `resource.programInstances[].programType` (verified against `ast.ts`), only gates `scopeType === "program"`, and doesn't interact with the pre-existing FUNCTION_BLOCK rule or the new `"configuration"` scope.
- :white_check_mark: **Tests** — checked out the actual PR branch in an isolated worktree (not just the diff) and ran the full suite: 2008 passed / 7 skipped across 79 files, matching the PR description. 15 new tests cover the location restriction, cross-scope duplicate detection, multi-instantiation rejection, and the new codegen array/accessors (including the STRUCPP_THREADED guard and the zero-globals placeholder case).
- :white_check_mark: **Type safety / lint** — `npm run typecheck` clean; `npm run lint` on the two changed files: 0 errors, only pre-existing `strict-boolean-expressions`-style warnings that match the file's existing conventions (not new patterns).
- :white_check_mark: **C++ compiles** — `tests/integration/cpp-compile.test.ts` (unchanged by this PR, but parametrized over `-DSTRUCPP_THREADED`) exercises the new code path through real g++, including the `void**` → `void *const*` qualification conversion in the new accessor — all 53 tests pass.
- :white_check_mark: **Scope matches description** — the fixture change (`IntersectionController.st` moving located I/O from `VAR_INPUT`/`VAR_OUTPUT` into `VAR`) and the `VAR_EXTERNAL AT` → plain `VAR_EXTERNAL` test fixup are both direct, minimal consequences of the new restriction, not unrelated cleanup.

### Code Improvement Findings
- `collectConfigurationLocatedVars()` only walks `ast.configurations[].varBlocks`; top-level `VAR_GLOBAL` blocks declared outside any `CONFIGURATION` (`ast.globalVarBlocks` — the "GVL file" path in `ast-builder.ts`) still bypass every rule this PR adds, the same class of gap this PR closes for configuration-scoped globals. Pre-existing, not a regression, but adjacent enough to this PR's stated goal that it's worth a follow-up. (inline comment on `analyzer.ts:792`)
- The new `strucpp_get_located_globals`/`strucpp_get_located_global_count` C ABI is a cross-repo wire contract with `openplc-runtime#161`; that repo isn't available locally so its shape couldn't be verified against its authoritative source per the wire-contract review step. Flagged for traceability, not blocking — the PR description states hardware-level end-to-end verification against the companion PR. (inline comment on `codegen.ts:5990`)

## Cross-Module Observations
None — single-package repo, changes are confined to `src/backend/codegen.ts` and `src/semantic/analyzer.ts` with matching test/fixture updates.

## Verdict
- **Approve**
- No blocking issues. Two non-blocking observations posted as inline comments (coverage gap for top-level GVL globals; unverifiable cross-repo ABI contract).

dcoutinho1328 (approved):

PR #205, opened 2026-08-10
feat: IEC 61131-3 structure and array initialization (forum report), plus six initialization fixes

thiagoralves:
Closes the forum report ["Initialization variable with user defined structure type"](https://edge.autonomylogic.com/forum/thread/initialization-variable-with-user-defined-structure-type). The user's syntax was correct IEC 61131-3 — the production simply wasn't implemented. Investigating it surfaced five more initialization defects, three of them silent wrong-value bugs.

## Structure initialization (Annex B.1.4.3)

```
test_ai_var : analog_input := (min_ai_scale := 4.0, max_ai_scale := 22.0);
```

Previously `error: Expected RParen, found :=`. Now supported in every declaration position: file-level and CONFIGURATION `VAR_GLOBAL`, PROGRAM / FUNCTION_BLOCK / FUNCTION / METHOD variables, STRUCT element defaults, inside array literals, and for function block instances (`t : TON := (PT := T#1s)` — the same production the standard uses for `fb_name_decl`).

Lowering lives in one place, `backend/struct-init-codegen.ts`, shared by `codegen.ts` and `type-codegen.ts`:

```cpp
strucpp::iec_struct_init<POINT>([](auto& v0) { v0.Y = 2.0; v0.X = 1.0; })
```

Not a braced aggregate initializer: elements may be written in any order and may be omitted, with an omitted element keeping the default from its own declaration — which C++17 cannot express (no designated initializers). The runtime helper default-constructs the value, applying each element's own default, and the lambda overwrites only what is named. Nested levels take their type from `decltype(v0.MEMBER)`, so library types and inline array members need no metadata lookup.

Also implements the type-level default forms of Annex B.1.3.3 (`Setpoint : REAL := 25.0;`, `Origin : Point := (x := 0.0);`), previously a parse error even for elementary types. A single post-build pass copies the default onto declarations without their own initializer, so every downstream consumer sees an ordinary initializer; it re-runs on the merged unit so a TYPE and its uses can live in different files.

## Array repetition (Annex B.1.4.3 `array_initial_elements`)

`[10(0)]`, `[3(1), 2(5)]`, `[7, 4(2), 9]`, and the bracket-less `:= 2(3), 2(4)`. The repeated value is a full expression, so it may be a structure initializer or a nested array literal. Gated on an integer immediately followed by `(`, which is never an expression in ST, so a function call as an array element is unaffected. Counts above 65536 throw rather than truncating — expansion is linear in the count.

## Nested array initializers, and 3D arrays

`[[1, 2, 3], [4, 5, 6]]` for a multi-dimensional array. Implemented in the runtime rather than by teaching codegen to detect ranks, so overload resolution handles it and the notation works in every scope automatically. Each level fills from its own lower bound, so a short inner list leaves the rest of that row at its default instead of shifting the next row up — the semantic difference from writing the values flat. `Array1D` also gained an element-typed list, so an array whose element is itself composite works (`ARRAY[0..1] OF Row`), and so does the nested-`Array1D` chain a 4+-dimensional array lowers to.

3D arrays were entirely unusable, not merely uninitialisable: `IEC_ARRAY_3D` had neither an initializer list nor `at()`, and codegen emits the bounds-checked `at()` for every subscript in a body. It is now at parity with 1D/2D.

## Function block array invocation

`units[0](step := 2.0)` did not parse — the statement was taken as an assignment target, which then demanded `:=`. A new `instanceCallStatement`, gated on `(` following the closing `]` directly, claims exactly the element invocation and leaves `arr[0].m(…)` to the existing rules. `FunctionCallExpression` gains an optional `instance`; the FB type resolves through the array's element type, and input assignment, the call, `VAR_IN_OUT` copy-back and `=>` capture all work against that expression unchanged, including variable and multi-dimensional indices.

## Bugs found and fixed along the way

Each was pre-existing; the first three were silent.

- **PROGRAM variable composite initializers were dropped.** `arr : ARRAY[0..2] OF INT := [1,2,3]` compiled clean and ran with a zero-filled array — the project model flattened initializers to a string via `expressionToString`, which had no case for array literals. It now carries the AST expression, which also removed the parallel string-lowering pass in `getDefaultValue` (based literals, digit separators, typed prefixes, time and calendar literals, strings), so declaration initializers and statement bodies can no longer disagree.
- **Array-literal defaults on a STRUCT element were dropped** to `{}` with no diagnostic.
- **STRING/WSTRING literals in the type generator were emitted unescaped**, so an embedded `"` closed the C++ string early. Latent until array defaults started emitting; OSCAT's HTML-entity tables hit it immediately. `translateIECString` moved to `codegen-utils` so a STRING literal lowers identically everywhere.
- **The debug pointer table indexed multi-dimensional arrays with chained `[i][j]`**, which `Array2D`/`Array3D` have no operator for — so any project with a multi-dimensional array and debug enabled failed to build (reported from an AVR build). A shared `formatArrayElementAccess` now owns the per-rank rule alongside `formatArrayType`.
- **File-level `VAR_GLOBAL` was invisible to `VAR_EXTERNAL`.** Such a reference now validates, including the type check, and is dropped from the pointer-plumbing list — file-level globals are plain file-scope storage the body already reaches by name, and a `GlobalVar<V>*` member would shadow the very global being referenced. The CONFIGURATION path is unchanged.
- **A TYPE naming a function block** (`AccumGrid : ARRAY[…] OF Accum`) emitted its alias before the POU forward declarations, so the element type was undeclared.

## Two initializer forms change shape

Both still valid C++ with the same value, and now identical to what the statement path emits: `INT#5` emits `static_cast<IEC_INT>(5)` instead of `5`, and `1.5E3` emits `1500.0`. The two assertions in `codegen-var-initializers.test.ts` that were pinned to the old string-lowering output are updated.

## Verification

**2132 tests passing** (86 files), up from 2006 — ~126 new across parser/AST, the shared lowering, codegen, semantics, and g++ integration. Coverage thresholds pass; `type-defaults.ts` at 100%.

The integration tests **run the binaries and assert values**, not just compilation — the only way to catch out-of-order elements landing on the wrong member, omitted elements losing their defaults, or a repetition group filling the wrong slots.

Also validated end-to-end through the OpenPLC Editor on a real project (44 assertions covering every form, plus the AVR `generated_debug.cpp` compiled for atmega2560), since the program and debug-table paths fail independently.

## Documented, not implemented

- `:= [10()]` — repetition with no value. No positional lowering in C++17, and neither matiec nor CODESYS accepts it.
- A *method* call on an array element, `units[0].M()` — needs the expression grammar's method-call lookahead to accept a subscripted object.
- No validation that an array initializer's shape or element count matches the declared dimensions, and none that a subscript count matches the rank. Ordinary mistakes there surface as C++ errors or silent truncation rather than ST diagnostics. Worth a follow-up.


thiagoralves:
Added a seventh commit: `feat(semantic): validate array initializer shape and subscript count`.

This closes the follow-up flagged in the description — the three mistakes that previously escaped to g++ or truncated silently now produce ST diagnostics with line and column:

```
ARRAY[0..2] OF INT := [1,2,3,4,5,6]              -> compiled, silently kept 1,2,3
ARRAY[0..1,0..1] OF INT := [[1,2,3],[4,5,6]]     -> compiled, silently dropped 3 and 6
ARRAY[0..1,0..1,0..2] OF INT := [[1,2,3],[4,5,6]] -> g++: no matching constructor
p[0,0] on a 3-dimensional array                   -> g++: no matching call to 'at'
```

Now:

```
fmt.st:3:29: error: Initializer for 'A' has 6 values but the array holds 3. The extra values would be discarded.
fmt.st:7:8: error: 'M' has 2 dimensions but is indexed with 1 index.
```

The subscript check walks the ordered access chain rather than the flat `subscripts` list, because only the chain distinguishes `a[0][1]` (two steps into an array of arrays) from `a[0,1]` (one two-index step into a 2D array) — the flat list reports 2 for both.

Every check is skipped rather than guessed at when the shape isn't statically known (`ARRAY[*]`, non-constant bounds, unresolved types, a dereference in the chain), so it can only add diagnostics for definite mistakes. 31 new tests, deliberately split between rejected and accepted cases — the accepted half is the point, since a false positive here would block valid code.

Suite: **2163 passing**, no net new lint warnings.

src/backend/codegen.ts
```diff
@@ -5560,95 +5557,109 @@ export class CodeGenerator {
...
+    // REF(); pointers via := ADR()/&.
+    if (
+      decl.referenceKind === "ref_to" ||
+      decl.referenceKind === "reference_to" ||
+      decl.referenceKind === "pointer_to"
+    ) {
+      return undefined;
     }
+    if (decl.initialValue) {
+      return this.generateInitializer(
```
marconetsf, line 5584: Dropping the string path loses precision on 64-bit integer initializers. `projectVarInitializer` now routes PROGRAM/VAR_GLOBAL initializers through `generateInitializer` → `generateExpression` → `formatIntegerLiteral`, whose plain-decimal branch is `return String(value)` where `value` is a JS `number` (`parseIECInteger` uses `parseInt`). The removed `lowerNumericInitializer` passed the *raw* digits through `iecBaseToCppLiteral` (which returns `raw.replace(/_/g, "")` for decimals), so it was exact.

```
PROGRAM Main VAR x : LINT := 9007199254740993; END_VAR
```

used to emit `X(9007199254740993)` and now emits `X(9007199254740992)` — a silently wrong constant. Either keep `rawValue` for plain decimals in `formatIntegerLiteral` (`return raw.replace(/_/g, "")`) or store integer literal values as `bigint`.

src/backend/codegen.ts
```diff
@@ -3580,6 +3610,13 @@ export class CodeGenerator {
         const elements = expr.elements.map((e) => this.generateExpression(e));
         return `{${elements.join(", ")}}`;
       }
+      case "StructInitializerExpression":
+        // A structure initializer needs the target's C++ type, which only the
+        // declaration site knows — see generateInitializer. Reaching here means
+        // one appeared where no type is available (it is not an expression IEC
+        // allows in a statement), so value-initialise rather than emit code that
+        // would not compile.
+        return "{}";
```
marconetsf: `case "StructInitializerExpression": … return "{}"` is a silent-wrong-value fallback, and it is reachable from real ST, not just from impossible positions. `generateExpression`'s `ArrayLiteralExpression` case maps elements through `generateExpression`, so `arr := [(x := 1.0), (x := 2.0)];` in a PROGRAM body emits `ARR = {{}, {}};` — every element zeroed, no diagnostic. The same happens for a structure initializer passed as a named FB argument and for SETUP vars in `test-main-gen.ts` (`emitExpression(decl.initialValue)`). Since a structure initializer in a non-declaration position is not legal IEC anyway, emit a `CompileError` (or a codegen warning) here instead of silently value-initialising.

src/backend/codegen.ts
thiagoralves: Confirmed and fixed in 070ebf8. I reproduced it against the base commit to be sure it was a regression and not just a pre-existing wart:

```
base fb7980f:  : X(9007199254740993), Y(18446744073709551615)     exact
this branch:   : X(9007199254740992), Y(9223372036854776000), Z(18446744073709552000)
```

Two things worse than the report:

- `ULINT := 18446744073709551615` emitted `18446744073709552000`, which is a **g++ hard error** (`integer literal is too large to be represented in any integer type`), not a silently wrong constant. `LINT := 9223372036854775807` overflows INT64_MAX and warns. Both are the types' own bounds — the values most likely to be written as sentinels.
- `type-codegen.literalToCpp` had **no `INT` case at all**; its `default:` branch carried the identical defect, so **STRUCT element defaults** were affected too, not only the paths you named.

**I went with the second of your two suggestions, deliberately.** Keeping `rawValue` for plain decimals reintroduces a defect this branch had accidentally *fixed* — a leading zero is an octal prefix in C++:

```
                base fb7980f     this branch
a : INT := 007;   A(007)            A(7)      ok by luck
b : INT := 0010;  B(0010)           B(10)     base was octal 8 — silently wrong
c : INT := 008;   C(008)            C(8)      base did not compile
```

So: `exactIntegerLiteralValue` parses the literal as a `bigint` (new `src/literal-utils.ts` — at the root because the analyzer needs it too), and `formatIntegerLiteral` lowers from that. Normalized rather than copied, which keeps both the 64-bit range and the leading-zero fix.

One detail worth flagging: a value above INT64_MAX now gets a **`ULL` suffix**. A C++ decimal literal is only ever given a *signed* type (C++17 [lex.icon]/3), so the exact digits alone still would not have compiled.

`type-codegen` now shares that one lowering with `codegen`, next to the STRING/array lowering they already shared.

While in there I also added a diagnostic for a literal wider than *any* IEC integer type, which previously reached g++ as an opaque error:

```
fmt.st:24:11: error: Integer literal '99999999999999999999999' is outside the range of every
                     IEC 61131-3 integer type (LINT holds -9223372036854775808 to
                     9223372036854775807, ULINT holds 0 to 18446744073709551615).
```

Zero false positives by construction — LINT/ULINT are the widest, so a value outside `[LINT_MIN, ULINT_MAX]` is wrong against every declared type. `LINT_MIN` parses as unary minus over a magnitude that exceeds LINT_MAX, so the unsigned bound is what keeps it accepted.

**Verification.** The new integration tests *run the binaries and compare values* — `9007199254740993` lowered to `...992` compiles perfectly, so only execution catches it. They build with `-Werror=implicitly-unsigned-literal -Werror=overflow`, since an unsuffixed ULINT_MAX is a GCC extension that warns rather than fails and the warning is the failure mode under test. Disabling the fix fails 14 of the 22 literal tests, three as g++ errors.

Also validated end-to-end through the CLI on a project mixing 64-bit accumulators with structure initializers, built under `-Wall -Wextra -Werror`:

```
accum=18446744073709551615 9007199254740993 -9223372036854775808 18446744073709551615 1000000 10
ticks=9223372036854775807
```

src/backend/codegen.ts
thiagoralves: Confirmed and fixed in 070ebf8. All three routes you named reproduce, each returning `success: true` with zero diagnostics:

```
arr := [(X := 1.0), (X := 2.0)];   ->  ARR = {{}, {}};
f(P := (X := 3.0));                ->  F.P = {};
s := (X := 1.0);                   ->  S = {};
```

One correction to the diagnosis, which makes it slightly worse rather than better: the elements are **not zeroed**. The generated `struct PT` carries an NSDMI, so `{}` value-initializes to the *type's declared defaults*. Built and ran it:

```
before: 9.000000 9.000000
after : 9.000000 9.000000     <- the written 1.0 / 2.0 vanished
```

Plausible-looking values rather than obvious zeros, so harder to spot in a real project.

Fixed in three parts, because the diagnostic alone would not have covered everything:

**1. Analyzer rejects the form in statement position** (`validateStructInitializerPlacement`), which gets line and column rather than a codegen-time message:

```
bad.st:22:12: error: A structure initializer '(NAME := value, ...)' is only valid as a
                     variable's initial value in a declaration, not inside a statement.
                     Assign the elements individually instead.
  22 |   bank := [(min_ai_scale := 1.0), (min_ai_scale := 2.0)];
     |            ^
```

The walk prunes at every initial value a declaration can carry — `VarDeclaration.initialValue` **and `TypeDeclaration.defaultValue`**. That second one is the type-level default (Annex B.1.3.3, `TYPE Origin : Point := (x := 0.0); END_TYPE`), a third legal position that is easy to miss because it is not a `VarDeclaration`; my first pass pruned only declarations and broke three existing tests on exactly that.

**2. `test-main-gen.ts:539` was a real bug, but not the one it looked like.** A TEST var block *is* a declaration, so `p : Point := (x := 1.0)` is legal there — it was going through `emitExpression` and emitting `POINT P = {};`. It now routes through `generateInitializer` like every other declaration and lowers properly:

```cpp
POINT P = strucpp::iec_struct_init<POINT>([](auto& v0) { v0.X = 1.5; });
```

**3. Codegen throws** instead of returning `{}`, as you suggested. It is now an internal invariant rather than a live path — `compile()` returns before codegen on any error, and the `.stst` flow runs `analyzeTestFile`, so nothing reaches it — but silent value-initialisation was the wrong failure mode for an invariant regardless.

14 new tests, deliberately split between rejected and accepted. The accepted half is the point: every position where IEC *does* allow the form — PROGRAM/FB/METHOD variables, STRUCT element defaults, type-level defaults, file-level `VAR_GLOBAL`, and structure initializers nested inside an array-literal initial value — has to keep working, or this blocks valid code.

**Unrelated find, not fixed here:** `generateSetupStruct` (`test-main-gen.ts:397`) drops SETUP member initializers *entirely* — `VAR k : INT := 5;` in a SETUP block emits `IEC_INT K;`. Pre-existing, affects scalars too, and needs a decision about whether the value belongs on the member or in `setup()`, so I left it out of this change rather than widen the scope. Happy to open a follow-up.

thiagoralves:
Added an eighth commit, 070ebf8, addressing both review comments from @marconetsf. Replies with the detail are in each thread; the short version:

**1. Integer literals lost precision above 2^53** ([thread](https://github.com/Autonomy-Logic/STruCpp/pull/205#discussion_r3757089755)) — valid, and a genuine regression against `fb7980f`, which I verified by building the base commit and diffing the output. Worse than reported in two ways: the LINT/ULINT *bounds* were a g++ **build failure**, not a silently wrong value, and `type-codegen` had no `INT` case at all, so STRUCT element defaults carried the same defect.

Fixed with the `bigint` route rather than the raw-passthrough one, deliberately: raw digits make a leading zero a C++ octal prefix, which this branch had accidentally fixed (`0010` was octal 8 on base, `008` did not compile). A value above INT64_MAX also needs a `ULL` suffix — a C++ decimal literal is only ever given a signed type.

**2. `StructInitializerExpression` → `"{}"` silent fallback** ([thread](https://github.com/Autonomy-Logic/STruCpp/pull/205#discussion_r3757089761)) — valid, all three routes reproduce with `success: true` and zero diagnostics. One correction: the elements are reset to the struct's *declared defaults* (NSDMI), not zeroed, so the wrong values look plausible.

Fixed in three parts: an analyzer diagnostic with line and column, `test-main-gen` routed through the proper initializer lowering (a TEST var is a declaration, so that one was a real bug rather than an impossible position), and codegen throwing instead of value-initialising.

**Also closed while in there:** a literal wider than any IEC integer type is now a diagnostic rather than an opaque g++ error.

```
fmt.st:24:11: error: Integer literal '99999999999999999999999' is outside the range of every
                     IEC 61131-3 integer type (LINT holds -9223372036854775808 to
                     9223372036854775807, ULINT holds 0 to 18446744073709551615).
```

**Verification.** Suite **2201 passing** (90 files), up from 2163 — 38 new, split between rejected and accepted cases. Typecheck clean, no net new lint warnings, coverage thresholds pass.

The integration tests **run the binaries and compare values**, since `9007199254740993` lowered to `...992` compiles perfectly and only execution catches it. They build with `-Werror=implicitly-unsigned-literal -Werror=overflow`, because an unsuffixed ULINT_MAX is a GCC extension that warns rather than fails. Reverting the fix fails 14 of the 22 literal tests, three as g++ errors.

Also validated end-to-end through the CLI on a project combining 64-bit accumulators with structure initializers in every legal position — builds under `-Wall -Wextra -Werror`, prints every value exactly, and the illegal forms produce five precisely located diagnostics.

**One unrelated find, not fixed here:** `generateSetupStruct` (`test-main-gen.ts:397`) drops SETUP member initializers entirely — `VAR k : INT := 5;` emits `IEC_INT K;`. Pre-existing, affects scalars too, and needs a call on whether the value belongs on the member or in `setup()`. Left out of this PR; worth a follow-up.

Gustavohsdp (approved):
## Approved — both review points closed, verified by mutation rather than by reading

I did not take the fix commit on trust: I broke each half of `070ebf8` on purpose to check the tests are load-bearing.

**64-bit integer precision.** Reverted `formatIntegerLiteral` to lower through the JS `number` again:

```
Test Files  2 failed | 88 passed (90)
      Tests  12 failed | 2189 passed
  ❯ tests/backend/codegen-integer-literal-exact.test.ts   (7 failed)
  ❯ tests/integration/integer-literal-exact-cpp.test.ts   (5 failed)
```

and among the failures, real hard compiler errors rather than only wrong values:

```
error: integer literal is too large to be represented in a signed integer type,
       interpreting as unsigned [-Werror,-Wimplicitly-unsigned-literal]
error: integer literal is too large to be represented in any integer type
AssertionError: expected '    : X(9007199254740992)' to contain 'X(9007199254740993)'
```

The integration tests are why this holds: they **compile and run the binary and compare the value**, which is the only way to catch `9007199254740993` becoming `...992` — that lowering compiles perfectly. Ran on clang here and `-Werror=implicitly-unsigned-literal` fired as intended.

Choosing the `bigint` normalisation over re-emitting the raw digits was the right call for a reason the original suggestion did not anticipate — the raw digits would have reintroduced the leading zero as a C++ octal prefix (`0010` → 8, `008` does not compile). Both stay fixed, and there is an executed test for it (`007 / 0010 / 008` → `7 10 8`). The typed-prefix path (`LINT#9007199254740993`) is covered too, and leaving based literals in their own notation is correct: a C++ hexadecimal literal *can* be given an unsigned type, a decimal cannot ([lex.icon]/3), which is exactly why only the decimal needs `ULL`.

**Structure initializer placement.** All three routes are now rejected in the analyzer with line and column. The correction to the diagnosis is worth keeping in the record: the elements were not zeroed — the generated struct carries an NSDMI, so `{}` value-initialised to the *type's declared defaults*, producing plausible values instead of obvious zeros. Splitting the new tests between rejected and accepted cases, and treating the accepted half as the point, is the right instinct — blocking valid code is the real risk in a check like this.

**Gates.** `ci.yml` triggers on `pull_request: branches: [main]` and this PR targets `development`, so CI never ran on it — GitHub reports "no checks reported on the branch". I ran them locally:

| gate | result |
|---|---|
| `lint` | **0 errors**, 506 warnings |
| `format:check` | clean |
| `typecheck` | clean |
| `test` | **2201 passing, 90 files, 0 failures** |
| `test:coverage` | 88.74% stmts / 85.94% branch / 96.09% funcs (thresholds 75) |
| `build` | ok, all `.stlib` archives rebuilt |
| `check:purity` | clean, 52 files |
| VS Code extension | **261 tests, 20 files**, bundle ok |

Against the merge-base `fb7980f`: 2008 → 2201 tests, 79 → 90 files across the PR.

---

### Two non-blocking notes

**1. One test passes for the wrong reason.** I removed the `TypeDeclaration.defaultValue` pruning from `validateStructInitializerPlacement` and **all 2201 tests still passed** — including `"accepts a type-level default (Annex B.1.3.3)"`. The branch is nonetheless load-bearing; without it this valid IEC is rejected:

```st
TYPE Origin : Point := (x := 0.0, y := 0.0); END_TYPE
PROGRAM Main VAR q : REAL := 1.0; END_VAR  q := q; END_PROGRAM
```

```
errors: ["A structure initializer '(NAME := value, ...)' is only valid as a
          variable's initial value in a declaration, ..."]
```

The existing test declares `VAR p : Origin;`, and the pass that copies a type-level default onto the declaration shares the *same expression object* — so the `VarDeclaration.initialValue` branch already prunes it. The case only depends on the `TypeDeclaration` branch when **no variable uses the type**. So the stated reason for that branch ("my first pass pruned only declarations and broke three existing tests") does not reproduce today, and nothing guards a future removal.

One line closes it: drop `VAR p : Origin;` from that test, or add a case with no consumer.

**2. Two small precision corrections**, neither material:

- "no net new lint warnings" is **+8** (parser +7, codegen +1, analyzer +1, debug-table-gen −1) — all warnings, zero errors, on a base that already carries 498.
- The `Library compilation failed: MATH.gvl.st:2:1: error: Symbol 'MATH' already defined` in the test log is **pre-existing** — it appears 5× on the base commit too. Not from this PR.

Separately, the CI trigger deserves its own ticket: no PR targeting `development` in this repo gets lint, typecheck, tests, coverage, extension or build checked before merge.

<sub>Review assisted by Claude Code.</sub>

Gustavohsdp (approved):

PR #213, opened 2026-08-15
fix(debug-table): mangle a member whose name matches its type, as codegen does

thiagoralves:
## What

A variable declared with its own type's name — `RunningLights : RunningLights`, which CODESYS allows and real projects use — is emitted by codegen as `RUNNINGLIGHTS_`, because GCC rejects a member that changes the meaning of its type name. The debug table addresses members by name and did not apply the same rule, so every entry for such an instance named a member that does not exist:

```
generated_debug.cpp:26: error: 'class strucpp::Program_MAIN' has no member named
'RUNNINGLIGHTS'; did you mean 'RUNNINGLIGHTS_'?
```

## Why it went unnoticed

`strucpp file.st` does not generate a debug table. The ST compiles, the program's C++ compiles, and the build dies in `generated_debug.cpp` — so this only reproduces in a full firmware build.

## Tests

Two regression tests: one asserting the mangled reference for `Motor : Motor`, one asserting a non-colliding member is untouched. Only the C++ expression changes; the trailing comment keeps the ST path the editor shows the user. Full suite: 2010 passing.

Verified end to end: a converted CODESYS project builds and runs on the OpenPLC simulator with this fix in place.

### DIFF
--- a/src/backend/codegen.ts
+++ b/src/backend/codegen.ts
@@ -5352,13 +5352,13 @@ export class CodeGenerator {
       if (arg.name) {
         // Named argument: assign directly
         this.emit(
-          `${indent}${instanceName}.${arg.name} = ${this.generateExpression(arg.value)};`,
+          `${indent}${instanceName}.${this.fbParamMemberName(arg.name, fbTypeName)} = ${this.generateExpression(arg.value)};`,
         );
       } else if (inputParamNames && positionalIndex < inputParamNames.length) {
         // Positional argument: map to VAR_INPUT by position
         const paramName = inputParamNames[positionalIndex];
         this.emit(
-          `${indent}${instanceName}.${paramName} = ${this.generateExpression(arg.value)};`,
+          `${indent}${instanceName}.${this.fbParamMemberName(paramName!, fbTypeName)} = ${this.generateExpression(arg.value)};`,
         );
         positionalIndex++;
       } else {
@@ -5398,7 +5398,7 @@ export class CodeGenerator {
         if (arg.name && inoutParams.has(arg.name.toUpperCase())) {
           this.emitCaptureToLvalue(
             arg.value,
-            `${instanceName}.${arg.name}`,
+            `${instanceName}.${this.fbParamMemberName(arg.name, fbTypeName)}`,
             indent,
           );
         }
@@ -5410,7 +5410,7 @@ export class CodeGenerator {
       if (arg.name && arg.isOutput) {
         this.emitCaptureToLvalue(
           arg.value,
-          `${instanceName}.${arg.name}`,
+          `${instanceName}.${this.fbParamMemberName(arg.name, fbTypeName)}`,
           indent,
         );
       }
@@ -5496,6 +5496,29 @@ export class CodeGenerator {
       this.memberMangledNames.set(name.toUpperCase(), mangled);
     }
     return mangled;
+  }
+
+  /**
+   * C++ member name for a parameter of the function block being invoked, by the
+   * same rule its declaration used (see member-mangling.ts).
+   *
+   * An FB whose input is named after its own type, or after an interface method
+   * it implements, is declared with a trailing underscore — so assigning through
+   * the bare name reaches a member that does not exist. Left alone when the FB
+   * type is unknown, which only disables the check.
+   */
+  private fbParamMemberName(
+    paramName: string,
+    fbTypeName: string | undefined,
+  ): string {
+    if (fbTypeName === undefined) return paramName;
+    return this.needsFieldMangling(
+      paramName,
+      this.resolveMemberType(fbTypeName, paramName),
+      fbTypeName,
+    )
+      ? `${paramName}_`
+      : paramName;
   }
 
   /**
--- a/src/backend/codegen.ts
+++ b/src/backend/codegen.ts
@@ -1548,11 +1548,7 @@ export class CodeGenerator {
         const cppType = this.mapTypeRefToCpp(decl.type);
         const tag = this.elaboratedTagIfShadowed(decl.type.name, fbMemberNames);
         for (const name of decl.names) {
-          const memberName = this.mangleMemberIfNeeded(
-            name,
-            cppType,
-            decl.type.name,
-          );
+          const memberName = this.mangleMemberIfNeeded(name, decl.type.name);
           this.emitHeaderLineDirective(decl.sourceSpan.startLine);
           const memberLine = this.currentHeaderLine;
           this.emitHeader(`    ${tag}${cppType} ${memberName};`);
@@ -1649,11 +1645,7 @@ export class CodeGenerator {
       for (const decl of block.declarations) {
         const cppType = this.mapTypeRefToCpp(decl.type);
         for (const name of decl.names) {
-          const memberName = this.mangleMemberIfNeeded(
-            name,
-            cppType,
-            decl.type.name,
-          );
+          const memberName = this.mangleMemberIfNeeded(name, decl.type.name);
           this.emitHeaderLineDirective(decl.sourceSpan.startLine);
           const memberLine = this.currentHeaderLine;
           if (decl.address) {
@@ -2177,11 +2169,7 @@ export class CodeGenerator {
             decl.type.name,
           );
           for (const name of decl.names) {
-            const memberName = this.mangleMemberIfNeeded(
-              name,
-              cppType,
-              decl.type.name,
-            );
+            const memberName = this.mangleMemberIfNeeded(name, decl.type.name);
             fbInits.push(`${memberName}(${initExpr})`);
           }
         }
@@ -2401,11 +2389,7 @@ export class CodeGenerator {
             ? { referenceKind: decl.referenceKind }
             : {}),
         });
-        const memberName = this.mangleMemberIfNeeded(
-          decl.name,
-          cppType,
-          decl.typeName,
-        );
+        const memberName = this.mangleMemberIfNeeded(decl.name, decl.typeName);
         // Map variable ST line → header member line
         const stLine = varSourceLines.get(decl.name);
         if (stLine !== undefined) {
@@ -2564,7 +2548,13 @@ export class CodeGenerator {
       for (const decl of prog.varDeclarations) {
         const initVal = this.projectVarInitializer(decl);
         if (initVal) {
-          inits.push(`${decl.name}(${initVal})`);
+          // Name the member as `generateProgramHeaderFromModel` declared it —
+          // `scale : Scale` is a collision (ST names are case-insensitive) and
+          // is declared `SCALE_`, so an initializer list naming `SCALE` does not
+          // compile. The FUNCTION_BLOCK constructor already does this.
+          inits.push(
+            `${this.mangleMemberIfNeeded(decl.name, decl.typeName)}(${initVal})`,
+          );
         }
       }
       // External globals: bind the pointer member to the canonical GlobalVar<V>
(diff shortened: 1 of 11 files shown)


marconetsf:
Diagnosis and the fix itself look right, and the comment explaining *why* GCC rejects the member is genuinely useful. Two notes on coverage.

### 1. `visitVarDecl` is one of four sites, and the only one fixed

`debug-table-gen.ts` builds a member expression in four places; the other three still concatenate the raw name:

| line | walker | covered |
|---|---|---|
| `:383` | interface vars of an FB, via `visitTypeRef` | no |
| `:399` | `varBlocks` of an FB, via `visitTypeRef` | no |
| `:422` | STRUCT fields, via `visitStructFields` | no |
| `:488` | `visitVarDecl` | **yes** |

Codegen mangles the uncovered ones by the same rule: `enterScope(fb.varBlocks)` at `codegen.ts:2166`/`:2187` with the declaration emitted through `mangleMemberIfNeeded` at `codegen.ts:1510` for FB members, and `type-codegen.ts:305-309` for struct fields.

So `Motor : Motor` inside a `PROGRAM` is fixed, but the same declaration inside a `FUNCTION_BLOCK` still emits `.MOTOR` against a member declared `MOTOR_` — the identical build failure. Same for a STRUCT field whose name matches its type. Worth checking against the converted CODESYS project that motivated this: if `RunningLights : RunningLights` lives inside an FB rather than the main program, this PR doesn't reach it.

### 2. `memberCppName` mirrors half of codegen's rule

`mangleMemberIfNeeded` (`codegen.ts:5497-5512`) mangles for two reasons:

1. the name collides with its own type name — replicated here
2. the name collides with an interface method name of the FB (`currentFBInterfaceMethods`) — not replicated

An FB implementing `IMotor` with a `Start` method and a `VAR Start : BOOL` gets `START_` from codegen and `.START` from the debug table — same class of failure.

Minor, and possibly deliberate: codegen also guards on `isUserDefinedType(stTypeName)` before mangling, while `memberCppName` compares names unconditionally. In practice that barely differs (`knownStructTypes` takes every `ast.types` entry, `knownFBTypes` includes library FBs), so if the simplification is intentional a one-line note would save the next reader the trip through `codegen.ts`.

### Root cause worth a follow-up

This mangling rule now exists in five copies with three different conditions:

- `codegen.ts:5491` `mangleMemberIfNeeded` — `isUserDefinedType` + interface methods
- `codegen.ts:5518` `needsFieldMangling` — same, via parent type
- `type-codegen.ts:305` — `!isElementaryType`
- `test-codegen.ts:166` — plain name compare
- `debug-table-gen.ts:469` (new) — plain name compare

A single shared `mangledMemberName(name, typeName, ctx)` would close the whole class instead of one case at a time — the duplication is what let this slip through in the first place.

---

On the plus side: the negative test (non-colliding member left alone) and stripping comments before the `not.toMatch` assertion are good calls — they stop the ST path in the trailing comment from producing a false green.

thiagoralves:
Thanks @marconetsf — all three points were valid, and the one flagged as minor turned out to be the most serious. Pushed as 67cb040 (the fixes) and 8ad26a3 (a seventh site found while retesting), with 23c08be merging `development` now that #205 has landed.

## 1. `visitVarDecl` is one of four sites — confirmed

Reproduced both uncovered paths as build failures, not just missing coverage:

```
FB member     -> g++: no member named 'MOTOR' in 'strucpp::RIG'
STRUCT field  -> g++: no member named 'INNER' in 'strucpp::RIG'
```

One refinement on the table: I traced which walker actually fires. **`:399` is the live path** — user-defined FBs surface members through `varBlocks` (their symbol's flat arrays are empty), so every `Motor : Motor` inside an FB went through it. **`:383` only ever sees library FBs.** I scanned all five bundled `.stlib` archives (588 sources, 605 KB) for an interface member named after its own type: zero hits. So that one is theoretical — fixed for symmetry, not urgency.

To your point about the CODESYS project: yes, had `RunningLights : RunningLights` lived inside an FB rather than the main program, the previous commit would not have reached it.

## 2. `memberCppName` mirrors half the rule — confirmed

The interface-method case is a build failure too, and a different one:

```
FUNCTION_BLOCK Drive IMPLEMENTS IMotor / VAR Start : BOOL

  header:  virtual IEC_BOOL START();   IEC_BOOL START_;
  table:   &g_config.INSTANCE0.D.START
  g++:     error: cannot create a non-constant pointer to member function
```

It takes the address of the *method*. Worth noting it reaches the table through `:399`, so points 1 and 2 had to be fixed together — extending `memberCppName` to the other sites without the interface condition would still have left this broken.

## 3. The `isUserDefinedType` divergence is not minor — it is a regression

> Minor, and possibly deliberate […] In practice that barely differs

It differs, and in the dangerous direction. Elementary type names are not reserved here (`VAR INT : INT;` compiles), so:

```st
VAR Time : TIME; Word : WORD; END_VAR
```

```
header:  IEC_TIME TIME;     <- codegen does NOT mangle
table:   .TIME_             <- name-only comparison mangles unconditionally
g++:     error: no member named 'TIME_' in 'strucpp::Program_MAIN'
```

I checked base `fb7980f`: it emitted `.TIME` / `.WORD` and **compiled clean**. So the previous commit broke a build that works on `development` today — the inverse of the bug it was fixing. `Time`, `Date`, `Word`, `Byte`, `Real` are all ordinary PLC variable names, and neither of the two tests covered it.

(The enum case is fine — `Color : Color` gets mangled by codegen through another path and the table agreed.)

## The root cause, closed

Took your suggestion: `src/backend/member-mangling.ts` owns `mangledMemberName(name, typeName, ctx)`, carrying both collisions. Callers supply what they can resolve — codegen its `known*Types` sets, the debug table the symbol tables and AST, `repl-main-gen` the AST via a shared `userDefinedTypeNames`. `type-codegen` gains an injectable `isUserDefinedType` so struct fields resolve FB and program names too (defaulting to its old "not elementary" behaviour standalone), and `test-codegen`'s inlined copy now calls `needsFieldMangling`.

One correction to your table: `test-codegen.ts:166` was not a plain name compare — it already used `isUserDefinedType` **plus** the interface check, the same condition as codegen. So it was five copies with two conditions plus `type-codegen`'s `!isElementaryType`, rather than three scattered ones. Your architectural point stands regardless, and closing it as a class is what surfaced the next two sites.

## Two more sites the review did not reach

**Sixth — `repl-main-gen.ts:250.`** Found by building the REPL end to end. Every `VarDescriptor` emits `&instance.MEMBER` with no mangling at all, so a project using the pattern would not link:

```
main.cpp:466: error: no member named 'RIG' in 'strucpp::Program_MAIN'
```

Note `rig : Rig` collides — ST names are case-insensitive, so this is easier to hit than the exact-case spelling suggests.

**Seventh — the PROGRAM constructor initializer list.** Found after merging #205. The header declares the member mangled; the initializer list named the raw name:

```
airange : AiRange := (hi := 22.0);

  header: AIRANGE AIRANGE_;
  .cpp:   : AIRANGE(strucpp::iec_struct_init<AIRANGE>(...))
  g++:    error: member initializer 'AIRANGE' does not name a non-static data member
```

Pre-existing on `development` — I verified it there on a clean worktree, without this branch. It was latent because it only bites when a colliding member *also* carries an initialiser, which #205's structure initializers make ordinary. The FUNCTION_BLOCK constructor already did this correctly; only the program-from-model path did not.

## Verification

**2234 passing** (92 files). Typecheck clean, no net new lint warnings, coverage thresholds pass.

`tests/integration/debug-table-cpp.test.ts` is new and **compiles the generated debug table with g++**, which nothing did before — exactly why this class reaches firmware builds silently, as the PR body says. `strucpp file.st` emits no table and `--build` does not include one. Every case above fails that test without these commits.

`tests/backend/member-mangling.test.ts` asserts the *agreement* rather than each emitter alone: declaration vs constructor initializer list, vs statement body, vs STRUCT field emission. Every bug here has been one emitter drifting from another, and only a cross-emitter assertion catches that.

End to end on the merged tree, on a project combining every collision with #205's multi-dimensional arrays, structure initializers and 64-bit literals:

- `strucpp --build` links and the REPL drives it — `COUNTER` 3 → `set 50` → 51, `WORD = 16#0007`, `EPOCH = 9007199254740993`.
- Linking the debug table into a binary confirms **all 32 entries address the member their ST path names**, including `RIG_.MOTOR_.RUN`, `DRIVE_.START_`, `COUNTS(1, 1)`, `CUBE(1, 1, 1)`, `AIRANGE_.HI`, and plain `TIME` / `WORD`.

The two rules compose on the same expression, which is the case I most wanted to see after the merge:

```
g_config.INSTANCE0.BANK(0, 0).MOTOR_.RUN     // #205's rank-aware subscript + the mangling
```

Also: the 2D/3D debug-table cases I had deferred are enabled now that #205's `formatArrayElementAccess` is in.

---

On your closing note — agreed, and it is what made the difference. Fixing this one case at a time would have missed the interface collision, the elementary-name regression, `repl-main-gen`, and the constructor initializer list. Only chasing it as a class turned them up.

marconetsf (approved):

thiagoralves:
Follow-up: **aa2ba1f** closes an eighth site, found by sweeping every remaining member-expression construction in the backend rather than waiting for the next build to fail.

## Eighth site — function-block invocation

Invoking an FB reaches its parameters through `instance.MEMBER` in four places — named input assignment, positional input assignment, `VAR_IN_OUT` copy-back, and `=>` output capture — none of which applied the rule:

```st
FUNCTION_BLOCK Sensor
VAR_INPUT Reading : Reading; END_VAR
```

```
header:  struct READING READING_;
.cpp:    S.READING = INP;
g++:     error: no member named 'READING' in 'strucpp::SENSOR'
```

Pre-existing, and **worse than the other seven in one respect**: it breaks the program's own C++, so it does not need debug enabled to bite. `strucpp file.st --build` fails on it directly.

`fbParamMemberName` resolves the parameter's declared type via `resolveMemberType` and defers to `needsFieldMangling` — the same "reach a member through a named owner" path the body and the debug table already use.

Verified with both collisions on one FB:

```
S.READING_ = INP;      // parameter named after its own type
S.ARM_ = true;         // parameter named after an implemented interface method
S.GAIN = 2.0;          // non-colliding sibling untouched
TALLY = S.ACC;         // VAR_IN_OUT copy-back
GOT = S.READING_;      // => capture
```

## Test gap it exposed

`debug-table-cpp.test.ts` only compiled `generated_debug.cpp`, so it could never have caught this — the invocation code lives in `generated.cpp`. It now compiles **both** in the same pass, which widens every case in that file, not just the new one.

## Where the eight sites stand

| # | Site | Found by | Breaks |
|---|---|---|---|
| 1 | `debug-table-gen` `visitVarDecl` | original PR | firmware build |
| 2 | `debug-table-gen` FB members (`:399`) | review pt.1 | firmware build |
| 3 | `debug-table-gen` STRUCT fields (`:422`) | review pt.1 | firmware build |
| 4 | `debug-table-gen` library FB interface (`:383`) | review pt.1 | theoretical — no bundled library hits it |
| 5 | interface-method half of the rule | review pt.2 | firmware build |
| 6 | `repl-main-gen` VarDescriptors | building the REPL | `--build` |
| 7 | PROGRAM constructor initializer list | merging #205 | any build, once the member has an initialiser |
| 8 | FB invocation (4 forms) | sweeping the remaining sites | any build |

Plus the over-mangling regression on elementary type names, which the review had flagged as minor.

Sites 2–8 were all pre-existing on `development`; 1 is what the PR originally fixed. Every emitter that names a member now routes through `member-mangling.ts`, and I re-grepped for stragglers — the remaining `.${...}` constructions in the backend are program-instance names (no collision possible: the member type carries a `Program_` prefix), method calls (the method keeps the name; it is the *variable* that yields), and array helpers.

## State

**2240 passing** (92 files). Typecheck clean, no net new lint warnings, coverage thresholds pass. `Time : TIME` re-verified on the pushed HEAD end to end — declarations, constructor list (`: TIME(5000000000LL), WORD(0xFF), REAL(1.5)`), and the debug table all unmangled, both artifacts compiling.

Nothing else outstanding on my side — this is ready for another look whenever you have time, @marconetsf.

PR #218, opened 2026-08-24
fix(debug): refuse writes and forces to CONSTANT variables

thiagoralves:
Phase 0 of NODE-94 (retain variables). Lands ahead of the rest because the editor's new Flags column makes this reachable on day one.

- **A `CONSTANT` was writable through the debugger.** It is emitted as a `const` C++ member, but the debug table reaches every leaf through a C-style `(void*)` cast that silently strips the qualifier — `static_cast` refuses the same conversion outright. Nothing then stopped `handle_set` / `handle_write` writing into a genuinely const object: UB, and a contradiction of what CONSTANT means. Nobody had hit it because no editor surface could create a const member yet.
- **Carries the qualifier through instead of losing it at the cast.** `Entry._pad` becomes `Entry.flags`, so the gate costs no flash and no RAM on any target — the byte was already there for alignment. `LEAF_FLAG_READONLY` is set by `debug-table-gen` and checked by both mutating paths, which return the new `STATUS_READ_ONLY` (0x86, next free after the licensing FCs). Reads are untouched; unforce is refused too, since a leaf that could never be forced has no force to clear.
- **`DebugMapV2` gains an optional `readOnly`** so the editor can hide the force control rather than offer an action that will be refused. Advisory only — the runtime enforces it, which is what keeps an older editor build or an OPC-UA client safe. **`version` stays at 2 on purpose:** the editor's `debug-parser.ts` rejects anything else outright, so a bump would break every editor pinned to an older strucpp release.

## Notes for review

Two decisions worth a second opinion:

1. **The flag is threaded as an explicit parameter, not a shared mutable.** More call sites touched, but a bit can be *cleared* partway down a subtree. Nothing does that today; RETAIN will (a `NON_RETAIN` member inside a `RETAIN` FB instance), and a mutable would leak the cleared value into the following sibling.
2. **The bit is OR-ed in at each declaring block, not only at program level.** Otherwise `g.SCALE` stays writable while the program's own `LIMIT` is gated — the same qualifier meaning different things by depth.

The AVR paths in `read_entry` assemble `Entry` field by field rather than copying the struct, so they read the flags byte explicitly. A missed read there returns 0, which reads as "writable" and would defeat the gate on exactly the targets with least room to spare.

## Verification

Propagation, across all four paths that can introduce the qualifier:

```
RO  G_MAX                  <- VAR_GLOBAL CONSTANT
    G_LIVE                 <- plain global, writable
RO  INST0.G.SCALE          <- VAR CONSTANT inside a function block
    INST0.G.READING        <- plain VAR in the same FB, writable
RO  INST0.LIMITS.A / .B    <- CONSTANT struct, both fields inherit
```

Enforcement, from a real binary rather than a syntax-only check (new test in `debug-table-cpp.test.ts` builds and runs it):

```
writable leaf : write=OK        force=OK
const leaf    : write=READ_ONLY force=READ_ONLY unforce=READ_ONLY
const leaf    : read=4 bytes, value=10 (declared 10)
const leaf    : storage after write attempts = 10 (intact)
```

Full suite: 92 files, 2247 passed / 7 skipped (up 7 from the 2240 baseline). No new lint warnings on the touched files.

## Follow-ups, not in this PR

- Editor side: hide the force control for `readOnly` leaves (openplc-editor / openplc-web, same phase-0 scope, separate repos).
- Phase 1: `NON_RETAIN` / `PERSISTENT` tokens, relax `VAR_INPUT` / `VAR_OUTPUT` RETAIN, reject RETAIN in `FUNCTION` / `METHOD` and on `VAR_TEMP`, fix the `IEC_COMPLIANCE.md` row.

Refs NODE-94

https://claude.ai/code/session_012FNp61926UEggtmsQPj3A3

### DIFF
--- a/src/runtime/include/debug_dispatch.hpp
+++ b/src/runtime/include/debug_dispatch.hpp
@@ -58,13 +58,14 @@ namespace strucpp { namespace debug {
 constexpr uint8_t STATUS_OK              = 0x7E;
 constexpr uint8_t STATUS_OUT_OF_BOUNDS   = 0x81;
 constexpr uint8_t STATUS_DATA_TOO_LARGE  = 0x82;
-// 0x83..0x85 are taken by the licensing FCs (MB_DEBUG_LIC_*); this is the
-// next free code. Returned when a write or force targets a leaf carrying
-// LEAF_FLAG_READONLY — an IEC CONSTANT. The refusal lives HERE, at the
-// bottom of the stack, so it holds for every caller: the editor's debugger,
-// an OPC-UA client, a plugin, or an older editor build that never learned to
-// hide the control.
-constexpr uint8_t STATUS_READ_ONLY       = 0x86;
+// 0x83..0x85 are taken by the licensing FCs (MB_DEBUG_LIC_*), and 0x86 by
+// PLC_SET_STATE's REFUSED_BY_SWITCH (see the editor's ModbusDebugResponse
+// enum) — this is the next actually-free code. Returned when a write or
+// force targets a leaf carrying LEAF_FLAG_READONLY — an IEC CONSTANT. The
+// refusal lives HERE, at the bottom of the stack, so it holds for every
+// caller: the editor's debugger, an OPC-UA client, a plugin, or an older
+// editor build that never learned to hide the control.
+constexpr uint8_t STATUS_READ_ONLY       = 0x87;
 
 // ---------------------------------------------------------------------------
 // Templated per-type helpers. One instantiation per IEC elementary type;
--- a/src/runtime/include/debug_dispatch.hpp
+++ b/src/runtime/include/debug_dispatch.hpp
@@ -58,6 +58,13 @@ namespace strucpp { namespace debug {
 constexpr uint8_t STATUS_OK              = 0x7E;
 constexpr uint8_t STATUS_OUT_OF_BOUNDS   = 0x81;
 constexpr uint8_t STATUS_DATA_TOO_LARGE  = 0x82;
+// 0x83..0x85 are taken by the licensing FCs (MB_DEBUG_LIC_*); this is the
+// next free code. Returned when a write or force targets a leaf carrying
+// LEAF_FLAG_READONLY — an IEC CONSTANT. The refusal lives HERE, at the
+// bottom of the stack, so it holds for every caller: the editor's debugger,
+// an OPC-UA client, a plugin, or an older editor build that never learned to
+// hide the control.
+constexpr uint8_t STATUS_READ_ONLY       = 0x86;
 
 // ---------------------------------------------------------------------------
 // Templated per-type helpers. One instantiation per IEC elementary type;
@@ -323,6 +330,13 @@ inline Entry read_entry(uint8_t arr, uint16_t elem) noexcept {
     const uint8_t* entry_addr = reinterpret_cast<const uint8_t*>(table_ptr) + elem * sizeof(Entry);
     uintptr_t ptr_val = pgm_read_word(entry_addr);
     uint8_t tag_val   = pgm_read_byte(entry_addr + sizeof(void*));
+    // `flags` sits immediately after `tag` — both uint8_t, no padding between
+    // them — so it is the byte after the tag. MUST be read here: the AVR paths
+    // assemble `out` field by field rather than copying the struct, and a
+    // missed flags read silently returns 0, which reads as "writable" and
+    // defeats the CONSTANT gate on exactly the targets with the least memory
+    // to spare for a second lookup.
+    out.flags = pgm_read_byte(entry_addr + sizeof(void*) + 1);
     out.ptr = reinterpret_cast<void*>(ptr_val);
     out.tag = tag_val;
 #elif defined(__AVR__)
@@ -335,6 +349,7 @@ inline Entry read_entry(uint8_t arr, uint16_t elem) noexcept {
     const uint8_t* entry_addr = reinterpret_cast<const uint8_t*>(table) + elem * sizeof(Entry);
     uintptr_t ptr_val = pgm_read_word(entry_addr);
     uint8_t tag_val   = pgm_read_byte(entry_addr + sizeof(void*));
+    out.flags = pgm_read_byte(entry_addr + sizeof(void*) + 1);
     out.ptr = reinterpret_cast<void*>(ptr_val);
     out.tag = tag_val;
 #else
@@ -354,6 +369,11 @@ inline uint8_t handle_set(uint8_t arr, uint16_t elem, bool forcing,
                           const uint8_t* bytes, uint16_t len) noexcept {
     Entry e = read_entry(arr, elem);
     if (!e.ptr || e.tag >= TAG__COUNT) return STATUS_OUT_OF_BOUNDS;
+
+    // A CONSTANT cannot be forced. Refused for BOTH directions: unforcing a
+    // leaf that could never be forced is a no-op, and returning OK for it
+    // would tell the caller a force had been cleared that never existed.
+    if (e.flags & LEAF_FLAG_READONLY) return STATUS_READ_ONLY;
 
     if (forcing) {
         uint8_t expected = type_ops[e.tag].size;
@@ -388,6 +408,10 @@ inline uint8_t handle_write(uint8_t arr, uint16_t elem,
                             const uint8_t* bytes, uint16_t len) noexcept {
     Entry e = read_entry(arr, elem);
     if (!e.ptr || e.tag >= TAG__COUNT) return STATUS_OUT_OF_BOUNDS;
+    // Same gate as handle_set. This is also the path the retain restore walk
+    // uses, so a CONSTANT can never be clobbered by a stale retained value
+    // either — constants come from the declaration, never from storage.
+    if (e.flags & LEAF_FLAG_READONLY) return STATUS_READ_ONLY;
     uint8_t expected = type_ops[e.tag].size;
     if (expected == 0) return STATUS_DATA_TOO_LARGE;  // string stub
     if (len < expected) return STATUS_DATA_TOO_LARGE;
--- a/src/backend/debug-table-gen.ts
+++ b/src/backend/debug-table-gen.ts
@@ -63,6 +63,33 @@ export const TAG = {
 
 export type TagName = keyof typeof TAG;
 
+// ---------------------------------------------------------------------------
+// Per-leaf flag bits — MUST match LEAF_FLAG_* in
+// runtime/include/debug_table.hpp. ABI: append only, never renumber.
+//
+// Carried down the leaf walk as an explicit parameter rather than a mutable
+// `currentFlags`, because a bit can be *cleared* partway down a subtree —
+// today nothing does, but the retain work adds exactly that (a NON_RETAIN
+// member inside a RETAIN function-block instance), and a shared mutable
+// would leak the cleared value into the following sibling.
+// ---------------------------------------------------------------------------
+export const LEAF_FLAG_READONLY = 1 << 0;
+
+/**
+ * Render an entry's flags byte as C++.
+ *
+ * Emits the named constant rather than a literal so `generated_debug.cpp`
+ * reads as intent — a reviewer scanning the table sees which leaves are
+ * gated without decoding a bitmask — and so a stale generated file fails to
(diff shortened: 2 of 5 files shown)


src/runtime/include/debug_dispatch.hpp
```diff
@@ -58,6 +58,13 @@ namespace strucpp { namespace debug {
 constexpr uint8_t STATUS_OK              = 0x7E;
 constexpr uint8_t STATUS_OUT_OF_BOUNDS   = 0x81;
 constexpr uint8_t STATUS_DATA_TOO_LARGE  = 0x82;
+// 0x83..0x85 are taken by the licensing FCs (MB_DEBUG_LIC_*); this is the
+// next free code. Returned when a write or force targets a leaf carrying
+// LEAF_FLAG_READONLY — an IEC CONSTANT. The refusal lives HERE, at the
+// bottom of the stack, so it holds for every caller: the editor's debugger,
+// an OPC-UA client, a plugin, or an older editor build that never learned to
+// hide the control.
+constexpr uint8_t STATUS_READ_ONLY       = 0x86;
```
dcoutinho1328: :x: **Wire-protocol status-code collision** — `STATUS_READ_ONLY = 0x86` collides with `REFUSED_BY_SWITCH = 0x86` in the editor's `ModbusDebugResponse` enum (`openplc-editor/src/backend/shared/simulator/types.ts:35`), which is the authoritative registry for this shared Modbus debug-status byte space (it already lists `SUCCESS=0x7E`, `ERROR_OUT_OF_BOUNDS=0x81`, `ERROR_OUT_OF_MEMORY=0x82`, `LIC_EMPTY=0x83`, `LIC_CORRUPT=0x84`, `LIC_UNSUPPORTED=0x85`, `REFUSED_BY_SWITCH=0x86`). That enum even has a regression test (`modbus-pdu-plc-control.test.ts:30`) asserting `0x86` is "clear of every other status code" in the registry. `REFUSED_BY_SWITCH` was added there on 2026-08-04/08-07 — well before this PR — so this isn't a race between concurrent PRs, the "next free code" comment here is just stale/wrong about what's already allocated.

Today this doesn't misfire visibly because the editor's `statusError()` in `modbus-pdu.ts` only special-cases `ERROR_OUT_OF_BOUNDS`/`ERROR_OUT_OF_MEMORY` and falls back to a generic "Unknown error code" for anything else — but the moment the editor-side follow-up (hiding the force control / surfacing a real "read-only" message, which this PR's own follow-up section calls out) adds a `READ_ONLY` entry to that same enum, it will collide with `REFUSED_BY_SWITCH` and break that uniqueness invariant.

Suggest: allocate `STATUS_READ_ONLY = 0x87` instead (the actual next free value in that registry), and fix the comment above it accordingly. Worth flagging to whoever owns the editor-side follow-up so `READ_ONLY` gets reserved at the same value in `ModbusDebugResponse` when they add it.

dcoutinho1328 (changes requested):
Verified the implementation logic against the runtime (both AVR read_entry branches, handle_set/handle_write ordering, Entry layout/sizeof) and against the codegen leaf-walk (every visitTypeRef/addLeaf call site threads `flags`, block-type restrictions checked in analyzer.ts confirm the FB-level and struct-field propagation are correct). Full suite (92 files / 2247 passed / 7 skipped) and the new C++ runtime-compile test pass. One blocking wire-protocol issue found by cross-checking the authoritative registry in the sibling `openplc-editor` repo.

dcoutinho1328:
# PR #218 Review

## Summary
Adds a `LEAF_FLAG_READONLY` bit to the debug entry table so the runtime refuses `handle_write`/`handle_set` (write, force, and unforce) against IEC `CONSTANT` leaves, closing a UB hole where the debug table's `(void*)` cast silently stripped `const` off genuinely-const C++ members. `debug-table-gen.ts` propagates the bit down every leaf-walk path (globals, program locals, FB members per-instance, struct fields, array elements); `DebugMapV2` gains an optional `readOnly` advisory field for the editor.

## Module: STruCpp core (backend codegen + runtime)

### Checklist Results
- :white_check_mark: **Correctness of the leaf-walk threading** — every `visitTypeRef`/`addLeaf`/`visitStructFields`/`visitVarDecl` call site was updated to thread `flags` (TypeScript's required-parameter typing plus a clean `typecheck` run make a missed call site essentially impossible here).
- :white_check_mark: **Block-type restrictions match the semantic layer** — `analyzer.ts`'s `validateVarModifiers` confirms only `VAR`/`VAR_GLOBAL`/`VAR_INPUT` can carry `CONSTANT` (`VAR_OUTPUT`/`VAR_IN_OUT` rejected), and the FB-member walk in `debug-table-gen.ts` checks `isConstant` for exactly the block types that can legally carry it.
- :white_check_mark: **Runtime gate ordering** — `handle_set`/`handle_write` check `LEAF_FLAG_READONLY` right after the `!e.ptr` bounds check, before the forcing/size branches, so write, force, *and* unforce are all refused for a CONSTANT leaf — matches the PR description and the new `debug-table-cpp.test.ts` runtime assertions.
- :white_check_mark: **AVR PROGMEM read paths** — both the `RAMPZ`-far and near-pointer branches of `read_entry` were updated to read `flags` at `entry_addr + sizeof(void*) + 1`, consistent with the `{ptr, tag, flags}` struct order; the non-AVR `#else` branch does a plain struct copy so it needs no change. (Neither AVR branch is exercised by CI — a pre-existing gap in this file, not something this PR introduces.)
- :white_check_mark: **`Entry` ABI unchanged** — `flags` replaces `_pad` in the same byte position, so `sizeof(Entry)` is unchanged on every target (verified by inspection of the struct layout and the accompanying comments).
- :x: **Wire-protocol status code collision** — see inline comment on `debug_dispatch.hpp:67`. `STATUS_READ_ONLY = 0x86` collides with `ModbusDebugResponse.REFUSED_BY_SWITCH = 0x86` in `openplc-editor`'s `src/backend/shared/simulator/types.ts` (the authoritative shared registry for this status-byte space, guarded by its own "clear of every other status code" regression test). That enum entry predates this PR by three weeks, so this is a stale "next free code" claim, not a race. Recommend `0x87`.
- :white_check_mark: **`version` correctly held at 2** — verified against the editor's actual `debug-parser.ts`, which rejects any `version !== 2`; the new `readOnly?: true` field is additive/optional and the editor's `DebugMapLeaf` interface simply doesn't read it yet (no strict schema validation that would reject the extra field), matching the PR's claim.

### Code Improvement Findings
- No duplication or extraction concerns beyond what's pre-existing in the file (e.g. the two AVR `read_entry` branches already shared a near-identical ptr/tag read pattern before this PR; this diff just extends that established pattern by one field rather than introducing new duplication).

## Cross-Module Observations
- The `LEAF_FLAG_READONLY` bit is IEC-semantic enforcement, not exclusively a C++ `const`/UB guard: function-block members are never emitted as C++ `const` in `generateFBHeaderDeclaration` (only `ast.globalVarBlocks` and `prog.varDeclarations` get a `const` qualifier), so gating `VAR CONSTANT` inside a function block prevents violating IEC semantics via the debugger even where there's no literal UB risk. This is consistent with the PR's own framing ("the same qualifier meaning different things by depth") and not a bug.
- Verified the wire-contract claims in the PR description (`version` pin, licensing FC allocation at 0x83-0x85) against the actual `openplc-editor` source rather than trusting the PR's own account — those checked out; only the 0x86 allocation did not.

## Verdict
- **Request Changes**
- Blocking: the `STATUS_READ_ONLY` value collides with an already-allocated code in the shared wire-protocol registry (`debug_dispatch.hpp:67`). Everything else in the diff is correct and well-tested.

dcoutinho1328 (approved):
**Incremental review:** reviewing commit 787e4d9..3a7244b (1 new commit). Previous commits were reviewed in an earlier pass.

Fix verified: `STATUS_READ_ONLY` moved from `0x86` to `0x87`. Re-checked against all three authoritative sources — openplc-editor's `ModbusDebugResponse` enum, `modbus_types.h`, and openplc-runtime's `debug_handler.c` — 0x87 is genuinely free in each. Comment updated to correctly note the 0x86 collision. Build, typecheck, lint, and the full test suite (including the C++ runtime-compile test) all pass. No other changes in this commit.

No outstanding findings.

dcoutinho1328:
# PR #218 Review (incremental)

> **Incremental review:** reviewing commits 787e4d9..3a7244b (1 new commit). Previous commits were reviewed in an earlier pass.

## Summary
Follow-up commit fixes the one blocking finding from the previous review pass.

## Module: STruCpp core (runtime)

### Checklist Results
- :white_check_mark: **Wire-protocol collision resolved** — `STATUS_READ_ONLY` moved from `0x86` (already claimed by `ModbusDebugResponse.REFUSED_BY_SWITCH` in openplc-editor) to `0x87`. Re-verified `0x87` is unallocated in all three authoritative sources: openplc-editor's `ModbusDebugResponse` enum, `modbus_types.h`, and openplc-runtime's `debug_handler.c`.
- :white_check_mark: **Comment accuracy** — the updated comment now correctly attributes `0x86` to `PLC_SET_STATE`'s `REFUSED_BY_SWITCH` rather than claiming it as free.
- :white_check_mark: **No regressions** — build, lint, typecheck, and the full test suite (92 files / 2247 passed / 7 skipped, including the C++ runtime-compile test) all pass. The tests reference the symbolic `STATUS_READ_ONLY` constant rather than the literal byte, so none needed updating.

### Code Improvement Findings
None.

## Cross-Module Observations
None beyond the previous pass.

## Verdict
- **Approve**
- No outstanding findings.

src/runtime/include/debug_dispatch.hpp
```diff
@@ -355,6 +371,11 @@ inline uint8_t handle_set(uint8_t arr, uint16_t elem, bool forcing,
     Entry e = read_entry(arr, elem);
     if (!e.ptr || e.tag >= TAG__COUNT) return STATUS_OUT_OF_BOUNDS;
 
+    // A CONSTANT cannot be forced. Refused for BOTH directions: unforcing a
+    // leaf that could never be forced is a no-op, and returning OK for it
+    // would tell the caller a force had been cleared that never existed.
```
thiagoralves, line 376: **correctness (low):** `handle_set` refuses *both* directions, so an unforce of a CONSTANT returns `STATUS_READ_ONLY` instead of being a no-op. Unforce was previously idempotent and total.

Scenario: an OpenPLC-side "clear all forces" sweep that walks every leaf calling `handle_set(arr, elem, false, …)` — the only bulk primitive this dispatch surface offers — now gets `0x86`/`0x87` on every CONSTANT leaf. A caller that treats a non-`STATUS_OK` reply as a hard failure aborts the sweep, leaving leaves after the first constant still forced.

The write/force refusals are clearly right; only the `forcing == false` direction is questionable. Worth confirming against the runtime's actual clear-forces flow before merging — the integration test locks this behavior in, so changing it later means changing a test that reads as intentional.

src/runtime/include/debug_dispatch.hpp
```diff
@@ -388,6 +409,10 @@ inline uint8_t handle_write(uint8_t arr, uint16_t elem,
                             const uint8_t* bytes, uint16_t len) noexcept {
     Entry e = read_entry(arr, elem);
     if (!e.ptr || e.tag >= TAG__COUNT) return STATUS_OUT_OF_BOUNDS;
```
thiagoralves, line 411: **correctness (low) — comment claims a guarantee this gate does not provide:** "This is also the path the retain restore walk uses, so a CONSTANT can never be clobbered by a stale retained value either."

That isn't true of the retain design on this PR's base branch. Retain save/restore goes through `RetainStorage::save/restore` with `RetainVarInfo{name, offsetof(...), sizeof(type)}` descriptors (`src/runtime/include/iec_retain.hpp`, emitted by `codegen.ts:generateRetainTable`) — a raw offset+size memory region. It never calls `handle_write` and never sees `Entry.flags`.

Scenario: `VAR RETAIN g : Gauge;` where `Gauge` declares `VAR CONSTANT SCALE`. The descriptor covers `sizeof(Gauge)`, so a restore memcpys over `SCALE` with whatever was persisted — exactly the clobber the comment says is impossible. (`RETAIN CONSTANT` on the *same* block is rejected by `analyzer.ts:1371`, which is what actually protects the simple case.)

Suggest dropping or narrowing the sentence so a future reader doesn't rely on it.

PR #222, opened 2026-08-25
feat(retain): CONSTANT, RETAIN and NON_RETAIN [NODE-94]

thiagoralves:
Implements the IEC 61131-3 variable qualifiers **CONSTANT**, **RETAIN** and **NON_RETAIN** in the compiler, the way CODESYS does. Lands with paired PRs on openplc-editor, openplc-web and openplc-runtime.

## What a qualifier now does

| Qualifier | Effect |
|---|---|
| `CONSTANT` | Leaf marked read-only; the debugger refuses writes and forces with status `0x87` |
| `RETAIN` | Leaf joins the retain blob; restored at start as a plain write |
| `NON_RETAIN` | Today's default, and the way to opt a member back out of a retained container |
| `PERSISTENT` | Folded into `RETAIN` — CODESYS's distinction has no analogue here |

Qualifiers sit on the var **block**, so a block can now carry a run of them (`MANY2` rather than `OPTION`), and `NON_RETAIN` *clears* the inherited bit rather than merely failing to set it. That is why flags travel down the leaf walk as a parameter and never as shared mutable state: a cleared bit must not leak into the next sibling.

`VAR_IN_OUT`, `VAR_TEMP` and `VAR_EXTERNAL` reject `RETAIN` — none of them owns storage. `VAR_INPUT` and `VAR_OUTPUT` accept it, matching CODESYS.

## The retain blob

Per-leaf granularity, addressed by the `(arrayIdx, elemIdx)` pairs the debug table already uses — no `offsetof`, no `sizeof`, nothing that depends on C++ layout. A 14-byte header carries magic, format, layout hash, payload length and crc32; the payload is one tightly packed slot per retained leaf.

The **layout hash** is FNV-1a over the ordered `path|typeTag` of every retained leaf — identity of the *layout*, deliberately not of the program. A body edit keeps retained values; adding, removing, retyping or reordering a retained variable invalidates them. Keying on the project MD5 instead would have discarded retained state on every unrelated edit.

Marshalling lives in `iec_retain.hpp`, parameterised by read/write/size function pointers so one implementation serves the AVR firmware and the Linux runtime's `.so`.

## Retained function blocks

A retained FB instance retains **everything it runs on**, not just its interface — including library blocks, whose internals the consuming compilation cannot see or even name. The library compiler flattens each block itself and ships the result in the manifest (`LibraryFBEntry.leaves`); **224 of 224** blocks across the five bundled archives produce a complete list, OSCAT's 172 included.

Locals surface only when the instance is retained, so the debugger keeps its black-box view everywhere else. An archive built before this format is **refused** with a message naming the block, rather than silently retaining half of it.

One walk serves both the debug table and the library compiler (`leaf-walker.ts`); they decide the C++ member name of every entry, and a disagreement there breaks the firmware build with nothing catching it earlier.

## Verification

`2282 passed | 7 skipped`, including compiled round-trips against the real runtime C++ that pack, wipe and restore — covering nested FBs, config globals, GVL structs, arrays, `NON_RETAIN` opt-outs, corrupt payloads, stale layouts and bad magic.

End-to-end on an SLM-RP4 and a P1AM-100. Two identical `TON`s differing only in `RETAIN`, both elapsed, then the program reloaded:

```
2s after reload   dwell(RETAIN): Q=True  ET=10s  STATE=2  |  loose: Q=False ET=3s160ms
```

## Note

`.stlib` archives are gitignored build artifacts. `prepack` runs `npm run build` and `files` ships `libs/`, so every release carries freshly-built archives.

https://claude.ai/code/session_012FNp61926UEggtmsQPj3A3


src/runtime/include/iec_retain.hpp
```diff
@@ -3,81 +3,300 @@
...
+                         SizeLeaf        size_of) noexcept {
+    if (debug::retain_var_count == 0) return LoadResult::Ok;  // nothing to do
+    if (blob == nullptr || len == 0) return LoadResult::Empty;
+    if (len < HEADER_SIZE) return LoadResult::Truncated;
+
+    if (get_u16(blob) != MAGIC) return LoadResult::BadMagic;
+    if (blob[2] != FORMAT_VERSION) return LoadResult::BadFormat;
+
+    const uint16_t payload = get_u16(blob + 8);
+    if (len < static_cast<size_t>(HEADER_SIZE) + payload) return LoadResult::Truncated;
```
dcoutinho1328: :x: **Correctness / memory safety — integer overflow bypasses the truncation check on 16-bit `size_t` targets (AVR)**

`len < static_cast<size_t>(HEADER_SIZE) + payload` can wrap on any platform where `size_t` is 16 bits — which is exactly avr-gcc's `size_t` (`unsigned int`), the target this header is explicitly vendored into per the file's own docstring ("the Arduino firmware ... vendor this directory").

`payload` (line 271) comes straight from 2 untrusted bytes inside the stored blob. If a corrupted or crafted blob sets `payload` close to 65535, `HEADER_SIZE (14) + payload` wraps mod 65536 to a small value, the truncation check on this line passes even though the real buffer (`len`) is far smaller, and the very next lines call `crc32(blob + HEADER_SIZE, payload, crc)` — reading up to ~64 KB past the end of `blob`. That defeats this function's own stated contract ("a corrupt or stale store degrades to a cold start rather than to plausible-looking garbage") on precisely the platform the contract exists for — a chip with far less than 64 KB of RAM to begin with.

Fix: avoid the overflowing addition — subtract instead, now that `len >= HEADER_SIZE` is already established a few lines up:
```cpp
if (len - HEADER_SIZE < payload) return LoadResult::Truncated;
```
`pack()`'s mirrored `total` computation (line 216) isn't reachable this way: `payload_size()` there reflects this program's own compiled retain set, not untrusted storage — this is specifically an `unpack()`-on-corrupted-input problem.

src/backend/debug-table-gen.ts
```diff
@@ -63,6 +63,91 @@ export const TAG = {
...
+// `currentFlags`, because a bit can be *cleared* partway down a subtree —
+// today nothing does, but the retain work adds exactly that (a NON_RETAIN
+// member inside a RETAIN function-block instance), and a shared mutable
+// would leak the cleared value into the following sibling.
+// ---------------------------------------------------------------------------
+export const LEAF_FLAG_READONLY = 1 << 0;
+/** Mirrors LEAF_FLAG_RETAIN in runtime/include/debug_table.hpp. */
+export const LEAF_FLAG_RETAIN = 1 << 1;
+
+/**
```
dcoutinho1328: :warning: **Doc comment orphaned from its function**

This JSDoc block ("Render an entry's flags byte as C++") documents `flagsLiteral`, but `flagsLiteral` is now defined 60+ lines below (around line 144). What sits directly under this comment is `applyBlockFlags`, which already has its own correct doc comment immediately below this one (lines 89-97). A reader scanning top-to-bottom sees two JSDoc blocks stacked above `applyBlockFlags` and no doc at all above `flagsLiteral`.

Looks like a reordering left this behind. Move it to sit directly above `flagsLiteral`.

src/backend/debug-table-gen.ts
```diff
@@ -63,6 +63,91 @@ export const TAG = {
...
+  block: { isConstant: boolean; isRetain: boolean; isNonRetain: boolean },
+): number {
+  let flags = inherited;
+  if (block.isConstant) flags |= LEAF_FLAG_READONLY;
+  if (block.isRetain) flags |= LEAF_FLAG_RETAIN;
+  if (block.isNonRetain) flags &= ~LEAF_FLAG_RETAIN;
+  return flags;
+}
+
+/**
```
dcoutinho1328, line 107: :warning: **Same issue — doc comment orphaned from its function**

This block ("FNV-1a (32-bit) over the retain layout...") documents `retainLayoutHashOf`, but that function is defined further below (around line 129). Directly under this comment sits another, correctly-placed doc comment for `RETAIN_HEADER_SIZE` (lines 122-126) followed by the constant itself. Same fix as the other comment on this file: move this block down to sit directly above `retainLayoutHashOf`.

src/backend/debug-table-gen.ts
```diff
@@ -647,16 +918,41 @@ export function generateDebugTable(
...
+    // fast path is the only thing it ever sees.
+    ...(retainVars.length > 0
+      ? {
+          retainVars: retainVars.map(({ arrayIdx, elemIdx, path, size }) => ({
+            arrayIdx,
+            elemIdx,
+            path,
+            size,
+          })),
+          retainLayoutHash: retainLayoutHashOf(retainVars),
```
dcoutinho1328: :warning: **Simplification — same hash computed twice**

`retainLayoutHashOf(retainVars)` is called here and again at line 926 for `renderCpp`'s parameter, over the same `retainVars` array both times. For a project with many retained leaves (the bundled libraries alone list thousands) this redoes the whole FNV-1a walk for no reason. Hoist into one local, e.g. `const retainLayoutHash = retainLayoutHashOf(retainVars);`, and pass that to both `renderCpp(...)` and the `debugMap` object.

tests/backend/debug-table-gen.test.ts
```diff
@@ -643,3 +644,550 @@ END_PROGRAM${CFG}`;
...
+
+    it("changes when retained variables are reordered", () => {
+      // Order is the blob's packing order, so a swap has to invalidate it.
+      expect(hashOf("VAR RETAIN a : DINT; b : INT; END_VAR", "a := 1;")).not.toBe(
+        hashOf("VAR RETAIN b : INT; a : DINT; END_VAR", "a := 1;"),
+      );
+    });
+  });
+
+describe("retain through a user-installed library", () => {
```
dcoutinho1328: :warning: **Nit — indentation implies the wrong nesting**

`describe("retain through a user-installed library", ...)` is written at column 0, reading as a top-level sibling of `describe("retain table", ...)` (opened at line 762). It's actually nested one level *inside* it — `describe("retain table")` isn't closed until line 1193, after this block. Its sibling `describe("layout hash", ...)` just above (line 1050) is correctly indented 2 spaces. Harmless for the tests themselves, just indent this block (and its `it`s) to match.

dcoutinho1328 (changes requested):
Reviewed the full diff against `development` (docs, frontend lexer/parser/AST, semantic analyzer, backend codegen/debug-table-gen, the runtime C++ headers, and the library compiler/loader/manifest), plus the new and updated tests. The feature itself — leaf-addressed retain table, the blob format, CONSTANT's read-only gate, and retaining a library FB's internal state — is well designed and heavily tested, including real g++-compiled round-trips. One memory-safety issue in the shared runtime blocks, the rest are small cleanups. Full breakdown in the summary comment.

dcoutinho1328:
# PR #222 Review

## Summary
Implements IEC 61131-3 `CONSTANT`, `RETAIN`, `NON_RETAIN` (and `PERSISTENT` folded into `RETAIN`) end to end: lexer/parser tokens, semantic validation with correct block-type and scope restrictions, a leaf-addressed retain table (replacing the old per-program `offsetof`-based `RetainVarInfo`), a `LEAF_FLAG_READONLY` runtime gate that refuses writes/forces to `CONSTANT` leaves, and retaining a library function block's *internal* state (not just its interface) via a new declarative `locals` field in the library manifest. Heavily tested, including real g++-compiled round-trips of the retain blob and the read-only gate.

## Module: Frontend (lexer / parser / AST)
### Checklist Results
- :white_check_mark: `NON_RETAIN` / `PERSISTENT` tokens added correctly, before `AT` doesn't matter but ordering vs. existing keyword/token lists looks consistent with the rest of the file.
- :white_check_mark: `MANY2` instead of `OPTION` for the qualifier set is the right call given qualifiers can combine (`RETAIN PERSISTENT`) — grammar stays permissive, semantics does the rejecting.
- :warning: (non-blocking, not filed as a fix-me) The PR description says "contradictory or repeated qualifiers are a semantic error." Contradictions are caught (`RETAIN`+`CONSTANT`, `RETAIN`+`NON_RETAIN`, `CONSTANT`+`NON_RETAIN`), but a literal repeat of the *same* qualifier (e.g. `VAR RETAIN RETAIN x : DINT;`) parses and compiles silently — harmless today since `isRetain`/`isConstant` are booleans, but it doesn't match the stated claim. Flagging as an FYI, not asking for a fix.

## Module: Semantic (analyzer / symbol-table)
### Checklist Results
- :white_check_mark: `hasInstanceState` threading is careful about not conflating "no instance" (FUNCTION/METHOD) with the existing `scopeType` check used for located-variable rules — the comment on why `scopeType` alone can't answer this is a good catch of a subtle trap.
- :white_check_mark: Contradiction checks ordered before block-type checks, with a clear rationale ("once two qualifiers disagree there is no single intent left to validate the rest against").
- :white_check_mark: `RETAIN` scope restrictions correctly relaxed to match IEC 61131-3 / CODESYS (`VAR_INPUT`/`VAR_OUTPUT` now allowed) — and the existing test suite was updated in place to assert the new, correct behavior rather than just adding new cases, so the old (wrong) restriction can't silently come back.
- :white_check_mark: `cppName` / `libraryName` additions to `VariableSymbol` / `FunctionBlockSymbol` are narrowly scoped and documented with the "why" (can't always re-derive mangling against the wrong compilation unit).

## Module: Backend (codegen / debug-table-gen)
### Checklist Results
- :white_check_mark: Old per-program `RetainVarInfo` / `offsetof`-based table cleanly removed from `codegen.ts`, including the doc comment explaining why the two `ProgramBase` vtable slots are kept (ABI pinning across the `.so` boundary) rather than deleted.
- :white_check_mark: Flags threaded as an explicit parameter down the leaf walk rather than mutable shared state — correctly reasoned given `NON_RETAIN` needs to *clear* a bit mid-subtree.
- :white_check_mark: Good test coverage for the walk: CONSTANT propagation through structs/arrays, RETAIN inheritance into nested FB instances, `NON_RETAIN` opt-out without leaking to siblings, library FB locals (both bundled and user-installed archives), layout-hash stability/invalidation.
### Code Improvement Findings
- :x: (inline comment) None here — the one blocking finding is in the runtime header, see below.
- :warning: (inline comments posted) Two JSDoc blocks got orphaned from their functions during what looks like a reorder (`flagsLiteral`'s doc now sits above `applyBlockFlags`; `retainLayoutHashOf`'s doc now sits above `RETAIN_HEADER_SIZE`) — see inline comments for exact lines.
- :warning: (inline comment posted) `retainLayoutHashOf(retainVars)` is computed twice for the same input (lines 926 and 945) — trivial to hoist into one local.

## Module: Runtime (C++ headers)
### Checklist Results
- :white_check_mark: `Entry.flags` repurposing the old `_pad` byte is a genuinely nice ABI-neutral way to add the gate — `sizeof(Entry)` is unchanged on every target, called out explicitly in the comment.
- :white_check_mark: Both AVR `read_entry` branches (far/near) were updated to read the new `flags` byte at the right offset; the non-AVR branch already gets it for free via the struct copy. Verified the offset math (`sizeof(void*) + 1`) against the actual AVR struct layout (4 bytes: 2-byte ptr, 1-byte tag, 1-byte flags) — correct.
- :white_check_mark: `handle_set` / `handle_write` both gate on `LEAF_FLAG_READONLY` before doing anything else, and the "why refuse unforce too" reasoning (an operation that never happened can't be reported as cleared) is sound.
- :white_check_mark: `pack()` / `unpack()` blob format: header layout, CRC coverage (header minus the CRC field itself, then payload), and "order is the packing order" addressing are all internally consistent, and `pack()`'s own bounds check (`cap < total`) is safe since `payload_size()` there reflects this program's own compiled, build-time-bounded retain set.
- :x: (inline comment posted) `unpack()`'s truncation check (`len < HEADER_SIZE + payload`) can integer-overflow on 16-bit-`size_t` platforms — i.e. AVR, the exact firmware target this header is vendored into — because `payload` is 2 untrusted bytes read straight from the stored blob. A corrupted/crafted blob can wrap the check and cause the following `crc32()` call to read up to ~64 KB past the real buffer. See the inline comment on `src/runtime/include/iec_retain.hpp:272` for the concrete scenario and a one-line fix (subtract instead of add, since `len >= HEADER_SIZE` is already established above).

## Module: Library (compiler / loader / manifest)
### Checklist Results
- :white_check_mark: `locals` on `LibraryFBEntry` is declarative (field descriptions, not pre-flattened leaves) — reuses the exact same walk the debug table already uses for user-defined FBs, so one walk covers both, and it degrades gracefully (with a named warning) for archives built before the field existed.
- :white_check_mark: `libraryName` marker replacing the "flat arrays populated" proxy for "is this FB a library one" is a real correctness fix in its own right — the proxy failed quietly in both directions per the comment, and this closes that.
- :white_check_mark: `buildLibraryTypeDefinition` giving library structs with exported fields a real `StructDefinition` (instead of the historical self-referential alias) is exactly what's needed for the debug/retain walk to descend into them, and is covered by the "user-installed library" test exercising a library-internal struct type.

## Cross-Module Observations
- The PR description references a `leaf-walker.ts` as the shared walk between the debug table and the library compiler; no such file exists in the diff or the repo — the shared walk logic actually lives inline in `debug-table-gen.ts`'s `visitTypeRef`/`memberCppName`. Purely a description inaccuracy, no code impact — not asking for a change, just flagging so it doesn't confuse someone searching for the file later.
- Test coverage is unusually strong for this kind of change: real g++-compiled round-trips for both the CONSTANT read-only gate and the retain blob pack/unpack cycle, not just AST/manifest assertions.

## Verdict
- **Request Changes**
- Blocking: the `unpack()` integer-overflow / out-of-bounds-read on 16-bit-`size_t` targets (`src/runtime/include/iec_retain.hpp:272`). Everything else is a small, optional cleanup.

thiagoralves:
## Review response — all five findings were already fixed

@dcoutinho1328 you fixed all of these yourself in `4197860`, so there is nothing from me on this branch. I verified each rather than take it on trust:

| Finding | State |
|---|---|
| `iec_retain.hpp:272` — 16-bit `size_t` overflow | Fixed — `len - HEADER_SIZE < payload` |
| `debug-table-gen.ts:80` — orphaned `flagsLiteral` doc | Fixed — doc now adjacent |
| `debug-table-gen.ts:107` — orphaned `retainLayoutHashOf` doc | Fixed — doc now adjacent |
| `debug-table-gen.ts:945` — hash computed twice | Fixed — one call site remains |
| test file — `describe` at column 0 | Fixed |

### On the overflow, because it deserves recording

Your analysis was right and the consequence is worse than "the check is wrong". I reproduced the arithmetic:

```c
/* AVR: size_t and unsigned int are both 16-bit */
14 + 65530 = 8            → len(100) < 8 is FALSE → check PASSES
                          → crc32(blob + 14, 65530, crc) reads 65530 bytes
                            off a 512-byte buffer

/* 32/64-bit host */
14 + 65530 = 65544        → len(100) < 65544 is TRUE → Truncated (safe)
```

So it was an out-of-bounds read of up to ~64 KB, reachable from two untrusted bytes in the stored blob, on a chip with a few KB of RAM — and unreachable on the Linux runtime, which is the one host where it would have been harmless anyway. Exactly the platform the "degrades to a cold start" contract exists for.

Worth a regression test for the `Truncated` path with a payload field near `0xFFFF`, since the round-trip suite currently only exercises a *short* blob (`HEADER_SIZE - 1`) and would not have caught this. Happy to add it here if you want it in this PR.

### Still blocking downstream, and not a code change

`openplc-editor` and `openplc-web` read `debugMap.retainBlobSize`, which exists only on this branch. Their `binary-versions.json` pins **v0.6.3**, and neither it nor **v0.6.4** carries `retainBlobSize` or `LEAF_FLAG_RETAIN` — so editor #1034 / web #691 compile locally only because a strucpp build is rsynced into `node_modules`.

**This PR needs to merge and release, and the pin bumped, before those two can go.** Flagged in their PR bodies too.

dcoutinho1328 (approved):

PR #224, opened 2026-08-26
feat(library): carry C/C++ and Python blocks in the .stlib

thiagoralves:
Enables [DOPE-585](https://autonomylogic.atlassian.net/browse/DOPE-585). Consumer PRs: Autonomy-Logic/openplc-web#699, Autonomy-Logic/openplc-editor#1042 — both depend on this and will need `binary-versions.json` bumped to the release that carries it.

## What this adds

A library may ship function blocks whose bodies are C/C++ or Python. STruC++ does not compile those bodies — it **transports** them. The archive carries the authored file verbatim, and the consuming toolchain lowers it at *its* build time against whatever native bridge that toolchain implements.

That distinction is the whole point. Compiling the bodies here would bake one fixed bridge ABI into every archive, so a bridge change would break every library published before it until each was rebuilt. Transporting the source keeps a published library working across bridge revisions. **It is worth guarding in review: making STruC++ compile these bodies looks like an obvious optimisation to anyone who does not know the history.**

What *is* read is the ST header every such file already carries — the `FUNCTION_BLOCK` line and its `VAR_*` blocks. That's ordinary ST, so the interface is recovered by projecting the file down to a header-only block and running it through the normal front end: one declaration parser, no second implementation to drift, and a bad declaration in a `.py` gets the same diagnostics an `.st` would. The body between the last `END_VAR` and the closing `END_FUNCTION_BLOCK` never reaches the parser.

Native blocks land in `manifest.functionBlocks` looking like any other, distinguished by `implementation: "cpp" | "python"` and pointing at their file via `sourceFile`. They emit no chunk.

## The four changes

1. **`--compile-lib <dir>` discovers `.cpp`/`.c`/`.cc`/`.cxx`/`.py`** alongside `.st`/`.il`, so an author drops `Block.py` in the folder next to `Other.st` and both end up in the archive. Discovery routes them into a separate bucket — feeding a `.py` to the ST parser was never going to work.
2. **The interface is parsed from the ST header**, exactly as for any other language; the body is skipped and carried as source.
3. **`--no-source` strips ST sources only.** Native bodies are exempt and must be: they have no chunk, so the source *is* the deliverable and an archive without it is unbuildable by anyone. A native block cannot be shipped closed-source in this format.
4. **The Node API accepts them too** — `compileStlib` routes by file extension, so a caller just includes `.cpp`/`.py` entries in the same `sources` array. That's the path the editor uses.

Plus: a library whose every input is native no longer fails. There's nothing for the compiler to do, so it isn't called. Previously this reached `compile()` with an empty translation unit and died on `No source files provided`, which is what made an all-native library unbuildable and is the production failure behind DOPE-585.

## Compatibility

Archives that contain no native blocks are **unchanged**, so the existing corpus and older editors are unaffected. An archive that *does* contain them is a newer artifact by definition: an older consumer finds no chunk for those blocks and fails loudly at compile rather than silently linking nothing.

A library with no inputs at all still fails, as before.

## Why this landed here rather than in the editors

An earlier iteration fixed this editor-side: the editor hand-built the archive strucpp refused to produce, and stamped non-standard `cppBlocks` / `pythonBlocks` fields onto it afterwards. Two problems with that, both of which this PR removes:

- The editor was duplicating this package's archive schema — `formatVersion`, the manifest field set, `chunks`, `sources` — making it the only thing holding an invariant it doesn't own. With web on 0.6.1 and editor on 0.6.3, that had to hold for two versions simultaneously.
- Those fields were non-standard, and `loadStlibFromString` normalises unknown top-level fields away. They survived only because the install path happens to persist the archive *text* rather than the parsed object. `manifest` and `sources` are first-class, so native blocks now survive every round-trip by construction.

The editors delete more than they gain: the synthesized archive, the post-hoc stamping, a `PLCVariable` projection in the archive, and an `originalPythonPous` sidecar all go away.

## Testing

- `tests/library/native-blocks.test.ts` — 25 new tests: extension mapping, source partitioning, header projection (including the rejections: no header, no variables, `PROGRAM`), all-native and mixed builds, chunk emission, the `--no-source` exemption, and loader round-trips.
- Full suite: **2240 passed, 7 skipped** — no regressions.
- CLI, by hand: all-native from a directory, mixed ST + native, and `--no-source`, each inspected for chunks / sources / manifest markers.

## Verified end-to-end on hardware

Built `nettest.stlib` (one C++ block, one Python block) with `strucpp --compile-lib`, installed it through the editor's real install path (`installFromFile` → `prepareStlibUpload` → `loadStlibFromString` → pool), created a consumer project that instantiates both, and compiled + uploaded to a device (`slm-rp4`, runtime v4.1.10, board `SLM-RP4`). The device compiled `pou_NETTEST__CPP_SCALE.cpp` and `pou_NETTEST__PY_OFFSET.cpp` and linked both into `new_libplc.so`.

Values read back over the debug channel:

| Variable | Expected | Device |
|---|---|---|
| `SCALER.SCALED` (C++ block) | 42 | **42** ✅ |
| `OFFSETTER.OUT_VAL` (Python block) | 142 | **142** ✅ |

Negative control — inputs changed to `RAW := 6, GAIN := 7, OFFSET := 9`, rebuilt and re-uploaded: **42** and **51**, both correct, so the blocks are computing on-device rather than returning anything stale.

Also driven from the openplc-web UI (`dev:local`, this branch's `dist` overlaid on its `node_modules/strucpp`): Build Library completed through the real flow including the avr-gcc verification stage — `Verification passed.` → `Library built successfully: build/nettest.stlib`.

https://claude.ai/code/session_018qaJbNUTA9kMLL9Jr5FjUU

[DOPE-585]: https://autonomylogic.atlassian.net/browse/DOPE-585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/library/library-compiler.ts
+++ b/src/library/library-compiler.ts
@@ -22,7 +22,6 @@ import {
 } from "./native-sources.js";
 import type {
   LibraryFBEntry,
-  LibraryFunctionEntry,
   LibraryManifest,
   LibraryVarType,
 } from "./library-manifest.js";
@@ -276,10 +275,11 @@ function compileNativeEntries(
   },
 ): {
   functionBlocks: LibraryFBEntry[];
-  functions: LibraryFunctionEntry[];
   errors: Array<{ message: string; file?: string; line?: number }>;
 } {
-  const empty = { functionBlocks: [], functions: [], errors: [] };
+  // Native blocks are always FUNCTION_BLOCKs — `projectNativeHeaderToSt`
+  // rejects FUNCTION and PROGRAM — so there is no function list to return.
+  const empty = { functionBlocks: [], errors: [] };
   if (native.length === 0) return empty;
 
   const projected: string[] = [];
@@ -346,9 +346,56 @@ function compileNativeEntries(
     functionBlocks: result.ast.functionBlocks.map((fb) =>
       tagNative(buildFBEntry(fb)),
     ),
-    functions: [],
     errors: [],
   };
+}
+
+/**
+ * Every exported name in a manifest that is claimed by more than one symbol.
+ *
+ * A library exports one thing per name. STruC++'s own duplicate detection only
+ * fires within a single kind and a single translation unit — two
+ * `FUNCTION_BLOCK Foo` in one compile is caught, but `FUNCTION Foo` beside
+ * `FUNCTION_BLOCK Foo` is not, and native headers compile in a separate unit
+ * so nothing there is compared against the ST at all.
+ *
+ * The result is a manifest with two entries under one name, no error, and a
+ * consumer picking whichever it happens to read last. Nothing downstream
+ * currently validates against these entries, so today it goes unnoticed — but
+ * a published `.stlib` is immutable, and the next reader that does validate
+ * inherits the ambiguity. Cheaper to refuse to emit it.
+ */
+function findDuplicateExports(manifest: LibraryManifest): string[] {
+  const kindsByName = new Map<string, string[]>();
+  const claim = (name: string, kind: string): void => {
+    const key = name.toUpperCase();
+    const kinds = kindsByName.get(key);
+    if (kinds) kinds.push(kind);
+    else kindsByName.set(key, [kind]);
+  };
+
+  for (const fn of manifest.functions) claim(fn.name, "function");
+  for (const fb of manifest.functionBlocks) {
+    claim(
+      fb.name,
+      fb.implementation
+        ? `${fb.implementation} function block`
+        : "function block",
+    );
+  }
+  for (const type of manifest.types) claim(type.name, "type");
+  for (const global of manifest.globals ?? [])
+    claim(global.name, "global variable");
+
+  const messages: string[] = [];
+  for (const [name, kinds] of kindsByName) {
+    if (kinds.length < 2) continue;
+    messages.push(
+      `"${name}" is exported ${kinds.length} times by this library (as ${kinds.join(", ")}). ` +
+        "A library exports one symbol per name — rename or remove one of them.",
+    );
+  }
+  return messages;
 }
 
 export function compileLibrary(
@@ -404,15 +451,25 @@ export function compileLibrary(
   // unit would fail, which is what used to make an all-native library
   // unbuildable.
   if (stSources.length === 0) {
+    const nativeOnlyManifest: LibraryManifest = {
+      ...emptyManifest(options),
+      functionBlocks: nativeEntries.functionBlocks,
+      headers: [`${options.name}.hpp`],
+      sourceFiles: sources.map((s) => s.fileName),
+    };
+    const duplicates = findDuplicateExports(nativeOnlyManifest);
+    if (duplicates.length > 0) {
+      return {
+        success: false,
+        manifest: emptyManifest(options),
+        headerCode: "",
+        cppCode: "",
+        errors: duplicates.map((message) => ({ message })),
+      };
+    }
     return {
       success: true,
-      manifest: {
-        ...emptyManifest(options),
-        functionBlocks: nativeEntries.functionBlocks,
-        functions: nativeEntries.functions,
-        headers: [`${options.name}.hpp`],
-        sourceFiles: sources.map((s) => s.fileName),
-      },
+      manifest: nativeOnlyManifest,
       headerCode: "",
       cppCode: "",
       chunks: [],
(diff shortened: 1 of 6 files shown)


marconetsf:
## Review — carrying C/C++ and Python blocks in the `.stlib`

**Verdict: changes requested.** Two blockers, both reproduced by running the code.

The architecture is right and it is well defended. STruC++ **transports** the native bodies instead of compiling them, and the reason — not freezing one bridge ABI into every published archive — is written exactly where someone would try to "optimise" it (`src/library/native-sources.ts:12-16`). I am guarding that: it should not be touched, and `expect(out.st).not.toContain("void loop")` (`tests/library/native-blocks.test.ts:133`) protects it executably rather than by comment alone.

What holds this up is not the architecture, it is the behaviour at the edges. There are two paths today where the library builds with `success: true` and hands back a wrong artifact, silently. A published `.stlib` is immutable in the field, so silence costs more here than a failure does.

**What I verified by running it** (isolated worktree at `956eb42`):

- the 25 new tests: **25 passed**
- `tsc --noEmit`: **clean**
- eslint on the five changed `src/` files: only pre-existing `no-console` warnings in `cli.ts`
- six probes I wrote against the edges — **four confirmed a defect**

**What I did not verify:** the full 2240-test suite. `scripts/rebuild-libs.mjs` (the vitest `globalSetup`) does not run from a worktree on this Windows box — `Only URLs with a scheme in: file, data, and node are supported ... Received protocol 'c:'` — which is pre-existing and unrelated to this diff. I take the 2240 as your report. I also did not repeat the hardware run.

---

## Blockers

### B1 — A `FUNCTION` in a native file is accepted, then silently dropped

`src/library/native-sources.ts:148-149` and `src/library/library-compiler.ts:344-352`

```ts
// native-sources.ts:148 — FUNCTION is explicitly accepted
const declaration =
  /^\s*(FUNCTION_BLOCK|FUNCTION|PROGRAM)\s+([A-Za-z_]\w*)/i.exec(body);
```

```ts
// library-compiler.ts:344-352 — but only functionBlocks is read, and functions is hardcoded []
return {
  functionBlocks: result.ast.functionBlocks.map((fb) => tagNative(buildFBEntry(fb))),
  functions: [],
  errors: [],
};
```

`projectNativeHeaderToSt` rejects `PROGRAM` with an explicit message (`native-sources.ts:162-167`) but accepts `FUNCTION`, projects its header and closes it with `END_FUNCTION`. The result lands in `result.ast.functions`, which nothing reads.

Reproduced with a `CPP_ADD.cpp` declaring `FUNCTION CPP_ADD : INT`:

```
success: true   errors: []
manifest.functions:      []
manifest.functionBlocks: []
manifest.sourceFiles: ["CPP_ADD.cpp"]
```

The library builds **successfully**, the file is transported in `sources`, and the block exists nowhere in the manifest. If it is the only input, the author gets an empty `.stlib` and a "Library built successfully". This is the same class of silent failure you (rightly) call the more dangerous of the two on the ABI line — just from the other side.

Cheapest honest fix is to reject it, matching how `PROGRAM` is handled:

```ts
if (kind === "FUNCTION") {
  return {
    fileName: source.fileName,
    message: `${source.fileName}: a native library block must be a FUNCTION_BLOCK, not a FUNCTION ("${name}"). A FUNCTION has no instance state for the consumer's native bridge to hold.`,
  };
}
```

Then `functions: []` and the `LibraryFunctionEntry` import (`library-compiler.ts:25`) become dead and can go. Supporting native `FUNCTION` properly is new scope; rejecting is what closes the hole today.

### B2 — A name shared by an ST block and a native block yields two manifest entries, with no error

`src/library/library-compiler.ts:528-533`

```ts
functionBlocks: [
  ...ast.functionBlocks.map((fb) =>
    tagDocumentation(tagCategory(buildFBEntry(fb), catByName), docByName),
  ),
  ...nativeEntries.functionBlocks,
],
```

The two lists are concatenated with no collision check. Native headers are compiled in a **separate** translation unit (`library-compiler.ts:316`), so the compiler never sees both names together and the duplicate detection that would fire within one TU never runs.

Reproduced with `DUP.st` and `DUP.cpp`, both declaring `FUNCTION_BLOCK DUP`:

```
success: true   errors: []
FBs: [{"name":"DUP","inputs":["A"]},{"name":"DUP","impl":"cpp","inputs":["Z"]}]
```

Two `DUP` entries with different interfaces — one with a chunk, one without. In the consumer, `registerLibraryArchives` (`src/backend/codegen.ts:766-767`) registers both in sequence and which one wins is an accident of ordering. Before this PR the case was impossible to construct; the PR opens it. It is not exotic either: it is an author moving a block from ST to C++ and forgetting to delete the `.st`.

Suggested fix — validate after `compileNativeEntries`, before assembling the manifest, and fail naming both files:

```ts
const nativeNames = new Set(nativeEntries.functionBlocks.map((fb) => fb.name.toUpperCase()));
const collisions = ast.functionBlocks.filter((fb) => nativeNames.has(fb.name.toUpperCase()));
if (collisions.length > 0) {
  return {
    success: false,
    manifest: emptyManifest(options),
    headerCode: "",
    cppCode: "",
    errors: collisions.map((fb) => ({
      message: `"${fb.name}" is declared both as an ST block and as a native block. A library exports one interface per name — remove one of the two files.`,
    })),
  };
}
```

Worth covering the native-vs-native collision too (`Foo.cpp` + `Foo.py`), which `languageByName` (`library-compiler.ts:295`) currently resolves by overwriting.

---



## What is well done

- The transport-not-compile decision is documented at the exact place a future reader would try to optimise it away, with the consequence spelled out (`native-sources.ts:12-16`), and the test at `native-blocks.test.ts:133` makes it a guarded invariant rather than a comment.
- Recovering the interface by projecting to a header-only block and reusing the ordinary front end is the right call: one declaration parser, no second implementation to drift, and a bad declaration in a `.py` gets the same diagnostics an `.st` would.
- The `--no-source` exemption is the correct one, and it is justified where a reader will actually meet it — both at `library-compiler.ts:649-655` and in the `StlibArchive.sources` docstring.
- `lastEndVarIndex`'s word boundary shows the false-positive class was considered; C1 is that the defence is scoped too widely, not that it is absent.
- Compatibility is handled carefully: archives with no native blocks serialize unchanged, and `nativeLanguageFor` is case-insensitive with the reason given (`native-sources.ts:52-56`).
- The tests cover the rejections (no header, no variables, `PROGRAM`) and round-trip through the real `loadStlibFromString`, and they drive the real `compileStlib` rather than a mock — acceptance criterion 6, on this side of the fence.
- The self-corrections on the Jira issue, including withdrawing a claimed gap, made this reviewable. That is worth saying.

## Previous requests

None — no prior reviews, inline comments or conversation on this PR. This is the first pass.

thiagoralves:
Thanks — both blockers were real and both are fixed in `6d3684b`. I reproduced each before changing anything, and two of the supporting claims did not hold up; details below so you can push back.

## B1 — fixed, and it was the important one

Reproduced exactly as you described:

```
success: true   errors: []
manifest.functions: []   manifest.functionBlocks: []
manifest.sourceFiles: ["CPP_ADD.cpp"]
```

Now a compile error naming the file and the reason:

> `CPP_ADD.cpp: a C/C++ library block must be a FUNCTION_BLOCK, not a FUNCTION ("CPP_ADD"). A FUNCTION has no instance state for the consumer's native bridge to hold. Declare it as a FUNCTION_BLOCK.`

Rejecting is right for a *native* function specifically — the bridge hands the body a pointer to per-instance storage and calls `setup` once before `loop` on every scan, which is function-block shape. Worth being explicit that this is **not** a statement about library functions in general: ST functions are part of libraries and must keep working. Verified they do, including alongside a native block:

```
ST FUNCTION alone            -> functions: ["ADD2:INT"]  chunks: ["ADD2"]
ST FUNCTION + native FB      -> functions: ["ADD2"]  FBs: ["CPP_SCALE/cpp"]  chunks: ["ADD2"]
```

`functions: []` and the `LibraryFunctionEntry` import are gone as you predicted.

**One half you didn't reach.** The same input hits a separate bug on the editor side: its orchestrator read native sources from a hardcoded `pous/function-blocks`, so a native FUNCTION died on a guessed path (`Could not read the source for CPP_ADD at pous/function-blocks/CPP_ADD.cpp`) *before* strucpp could explain itself. The directory now follows the POU type, so your message is what the author actually sees. Fixed in the consumer PRs.

## B2 — fixed, wider than you proposed

Reproduced: two `DUP` entries, `success: true`, no error.

I went broader than a native-vs-ST check, because probing showed the gap isn't specific to this feature — strucpp's duplicate detection fires only within a single kind:

```
ST  FUNCTION FOO + FUNCTION_BLOCK FOO, same TU   ACCEPTS
ST  TYPE FOO     + FUNCTION_BLOCK FOO, same TU   ACCEPTS
native FB FOO + ST FUNCTION / TYPE / GLOBAL FOO  ACCEPTS
```

So `findDuplicateExports` now runs over the whole assembled manifest — functions, function blocks, types, globals, case-insensitively — and fails naming the symbol and both kinds. All seven shapes now reject; the two the front end already caught keep their original messages. A library with distinct names is unaffected, and the full suite is **2265 passed** with every bundled library rebuilding through the new check.

**Where I disagree:** the consumer-side consequence doesn't reproduce. Against a library containing *only* the ST block:

```
d(Z := TRUE)   -> success=true     <- interface not in this library at all
d(NOPE := 1)   -> success=true     <- name that exists nowhere
d(5)           -> success=true
```

Library FB argument names aren't validated anywhere, so `fbInputParams` last-wins has nothing observable to act on. Editor-side the two entries surface under *different* names (`DUP` and `duplib__DUP`) because the graft prefixes native blocks, so the picker doesn't collide either. I still fixed it, and I think it's worth fixing — an immutable artifact shouldn't carry an ambiguity for the next reader that *does* validate — but as archive hygiene rather than the correctness hole the write-up implies.

## Two things to correct

**The native-vs-native sub-claim is wrong.** You suggested `languageByName` "resolves by overwriting". It never gets there: both headers share one synthetic translation unit, so the front end's own detection fires first — `FOO.cpp` + `FOO.py` gave `Duplicate function block declaration: FOO` and `success: false` *before* this change. Already correct, already loud. I added a test to lock it in.

**C1 doesn't exist in the comment.** It's referenced once — *"C1 is that the defence is scoped too widely"* — but the item itself never made it in. I inferred it (last-`END_VAR` slicing running into a body that contains the token) and confirmed it fails loudly with a confusing parser message rather than producing a wrong artifact. Deliberately not addressed: the editor's `findLastEndVarIndex` uses the identical heuristic, so tightening one side alone would make the two disagree about where a header ends, which is worse than the shared wart. Happy to take it as a paired change if you still want it — please post the missing item.

Also: I couldn't reconcile "four of six probes confirmed a defect". I found two real defects, one non-defect (native-vs-native) and one lower-severity issue (C1). If two more probes found something, I'd like to see them.

I ran the full suite you couldn't — **2265 passed, 7 skipped** — and did not repeat the hardware run for this follow-up, since neither fix touches the transport path that the run exercised.

marconetsf (approved):

PR #230, opened 2026-09-01
fix(lsp): struct-field symbol resolution and member access past an array subscript (RTOP-247, RTOP-248)

JoaoGSP:
Closes RTOP-247 and RTOP-248. Two defects in the same resolver layer, shipped together so the consuming repos pay a single `binary-versions.json` bump.

## RTOP-247 — a STRUCT field resolves to a global symbol

Struct fields are `VarDeclaration` nodes, the same node kind as a POU's VAR block declarations, so both reach the same arm of `resolveSymbolAtPosition`. Inside a POU the analyzer has already defined the local and it wins the lookup. Inside `TYPE ... END_TYPE` there is no such symbol — `findEnclosingPOU` has no TYPE branch, so the scope is `globalScope`, and field names are never defined as symbols at all. `globalScope.lookup("Motor")` answers with whatever global shares the name.

A declaration position must not borrow a symbol from the global scope, so a field now returns its node with no symbol.

**This is wider than the card described.** The card called it a misleading tooltip. Five call sites share this resolver:

| surface | before |
| --- | --- |
| hover | showed the function block's signature |
| go-to-definition | navigated to the function block |
| **rename** | **renamed the FUNCTION BLOCK and every reference to it** |
| find-all-references | listed the block's references |
| `isWrapped` server query | misclassified the field |

The rename case is data loss, not cosmetics, and is pinned by its own tests.

It is also not function-block only — `lookup` returns any global symbol kind, so a field named after a FUNCTION mis-resolves identically. Tested.

## RTOP-248 — member access past an array subscript

`myArray[i].` offered the generic body-completion list — keywords plus every global symbol — instead of the element type's members. Worth noting for anyone reproducing: the symptom is a list of the *wrong* things, not an empty list.

Two causes, both in the completion stack:

- `getCursorContext`'s dot-access regex matched only bare identifier chains, so a prefix ending in `]` never became a dot-access context.
- The chain resolver walked field steps only. Segments now carry their subscript count and each peels one array level through the compiler's own `resolveArrayElementType` (`src/semantic/type-utils.ts`), which already understands both `__INLINE_ARRAY_*` synthetic names and named array TYPEs — so the synthetic-name obstacle the card flagged disappears with it rather than needing separate handling.

Covered: inline arrays, named array TYPEs, multi-dimensional arrays, an array reached through a struct field, and arrays of elementary types (correctly no members).

**Scope correction worth a reviewer's attention.** RTOP-248 also claims hover and go-to-definition are broken past a subscript. They are not: both resolve to the base variable there, and they do exactly the same for a plain non-array struct. That is the existing behaviour of those surfaces, not a defect of arrays. Measured before touching anything:

| | `plain.x` (baseline) | `pts[i].x` |
| --- | --- | --- |
| completion after `.` | `x`, `y` | 142 keyword/global items |
| hover on `x` | base variable | base variable |
| go-to-def on `x` | base variable's declaration | base variable's declaration |

Making them answer per-field would change both paths and runs into RTOP-247's finding that fields are not symbols at all. Left alone deliberately.

## Tests

14 tests across `resolve-symbol`, `hover`, `definition`, `rename` and `completion`, all red-green verified — stashing each fix fails exactly its own tests and nothing else. Full extension suite **277/277**; root and extension `tsc` clean.

Lint: the extension is outside CI's lint and format jobs and carries 227 pre-existing problems, so I compared per file. `resolve-symbol.ts` is back at its baseline (an earlier draft added an unnecessary type assertion, removed). `completion.ts` went **16 → 13 problems**, since the explicit `=== undefined` checks also cleared some pre-existing warnings. `cursor-context.ts` is clean either way. Prettier flags three of these files identically with and without the change; formatting them would add unrelated churn.

## Validation

Built and wired into the OpenPLC editor against a project carrying every shape: a `SensorData` struct with fields named after a FUNCTION_BLOCK and a FUNCTION, an inline array of a struct, a named array TYPE, a multi-dimensional array, an array behind a struct field, and an array of REAL. Both fixes confirmed live.

## Release note

Both consuming repos (`openplc-web`, `openplc-editor`) need a `binary-versions.json` bump once this is released — one bump for both cards, which is why they ship together.

https://claude.ai/code/session_011JAMx8mRf2ig4YFfs2gmsM

### DIFF
--- a/vscode-extension/server/src/completion.ts
+++ b/vscode-extension/server/src/completion.ts
@@ -259,13 +259,36 @@ interface ChainSegment {
   subscripts: number;
 }
 
-/** Split "a.b[0].c" into its links, counting the subscripts on each. */
+/**
+ * Split "a.b[0].c" into its links, counting the subscripts on each. Dots inside
+ * a subscript belong to the index expression (`arr[s.k]`), not to the chain, so
+ * the split only happens at bracket depth 0.
+ */
 function parseChainSegments(prefixExpr: string): ChainSegment[] {
-  return prefixExpr.split(".").map((raw) => {
-    const name = raw.replace(/\[[^\]]*\]/g, "");
-    const subscripts = raw.length === name.length ? 0 : (raw.match(/\[[^\]]*\]/g) ?? []).length;
-    return { name, subscripts };
-  });
+  const segments: ChainSegment[] = [];
+  let name = "";
+  let subscripts = 0;
+  let depth = 0;
+
+  for (const ch of prefixExpr) {
+    if (ch === "[") {
+      if (depth === 0) subscripts++;
+      depth++;
+    } else if (ch === "]") {
+      if (depth > 0) depth--;
+    } else if (depth === 0) {
+      if (ch === ".") {
+        segments.push({ name, subscripts });
+        name = "";
+        subscripts = 0;
+      } else {
+        name += ch;
+      }
+    }
+  }
+  segments.push({ name, subscripts });
+
+  return segments;
 }
 
 /** Peel one array level per subscript. Undefined once a step is not an array. */
--- a/vscode-extension/server/src/completion.ts
+++ b/vscode-extension/server/src/completion.ts
@@ -21,8 +21,9 @@ import type {
   FunctionBlockType,
   SymbolTables,
   Scope,
+  CompilationUnit,
 } from "strucpp";
-import { ELEMENTARY_TYPES, typeName } from "strucpp";
+import { ELEMENTARY_TYPES, resolveArrayElementType, typeName } from "strucpp";
 import { getCursorContext } from "./cursor-context.js";
 import { getScopeForContext } from "./resolve-symbol.js";
 import { isTestFile, extractTestVarDeclarations } from "../../shared/test-utils.js";
@@ -217,30 +218,31 @@ function getDotAccessCompletions(
   const scope = getScopeForContext(symbolTables, pouScope);
   if (!scope) return [];
 
-  // Parse the chain: "a.b.c" → resolve segment by segment
-  const segments = prefixExpr.split(".");
-  const resolvedType = resolveChainType(segments, scope, symbolTables);
+  // Parse the chain: "a.b[0].c" → resolve segment by segment
+  const segments = parseChainSegments(prefixExpr);
+  const resolvedType = resolveChainType(segments, scope, symbolTables, analysis.ast);
   if (resolvedType) return getMembersForType(resolvedType, symbolTables);
 
   // For test files, try resolving via locally declared variable types
   if (testSource) {
     const testVars = extractTestVarDeclarations(stripCommentsAndStrings(testSource));
-    const varType = testVars.get(segments[0].toUpperCase());
+    const varType = testVars.get(segments[0].name.toUpperCase());
     if (varType) {
       // Walk remaining segments through type chain
-      let currentTypeName = varType;
-      for (let i = 1; i < segments.length; i++) {
-        const nextType = resolveMemberType(currentTypeName, segments[i], symbolTables);
-        if (!nextType) return [];
-        currentTypeName = nextType;
+      let currentTypeName = unwrapSubscripts(varType, segments[0].subscripts, analysis.ast);
+      for (let i = 1; i < segments.length && currentTypeName !== undefined; i++) {
+        const nextType = resolveMemberType(currentTypeName, segments[i].name, symbolTables);
+        if (nextType === undefined) return [];
+        currentTypeName = unwrapSubscripts(nextType, segments[i].subscripts, analysis.ast);
       }
+      if (currentTypeName === undefined) return [];
       const typeInfo = resolveTypeName(currentTypeName, symbolTables);
       if (typeInfo) return getMembersForType(typeInfo, symbolTables);
     }
   }
 
   // Fallback: first segment may be a type name (e.g., EnumType.MEMBER)
-  const typeInfo = resolveTypeName(segments[0], symbolTables);
+  const typeInfo = resolveTypeName(segments[0].name, symbolTables);
   if (typeInfo) return getMembersForType(typeInfo, symbolTables);
 
   return [];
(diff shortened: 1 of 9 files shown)


vscode-extension/server/src/completion.ts
```diff
@@ -251,30 +253,65 @@ interface ResolvedTypeInfo {
...
+  typeName: string,
+  subscripts: number,
+  ast?: CompilationUnit,
+): string | undefined {
+  if (subscripts === 0) return typeName;
+  if (!ast) return undefined;
+  let current: string | undefined = typeName;
+  for (let i = 0; i < subscripts && current !== undefined; i++) {
+    current = resolveArrayElementType(current, ast);
+  }
```
thiagoralves, line 305: `resolveArrayElementType` only understands `__INLINE_ARRAY_*` and a direct `ArrayDefinition`, so two array spellings the compiler fully supports now resolve to `undefined` here and the user gets **zero** completions after the dot (before this PR they at least got the generic body list):

**1. A TYPE alias to a named array type**
```st
TYPE PointArray : ARRAY [0..9] OF Point; END_TYPE
TYPE AliasArr   : PointArray;            END_TYPE
VAR ali : AliasArr; END_VAR
```
`ali[i].` → `[]`. `resolveArrayElementType("AliasArr")` hits the `TypeReference` definition and falls off the end of the loop. Meanwhile `r := ali[i].x;` compiles clean and codegens `R = ALI.at(I).X;`.

**2. A variable-length array (`ARRAY [*]`, i.e. a `VAR_IN_OUT` array)**
```st
FUNCTION_BLOCK Sum
VAR_IN_OUT vals : ARRAY [*] OF Point; END_VAR
```
`vals[i].` → `[]`. The AST builder names these `__VLA_1D_POINT` (`ast-builder.ts:1606`), which the `__INLINE_ARRAY_` prefix check doesn't match. `r := vals[i].x;` also compiles clean and codegens `R = VALS.at(I).X;`.

Both measured on this branch. `resolveArrayShapeByName` in the same module (`src/semantic/type-utils.ts`) already walks `TypeReference` alias chains — following aliases before the call, and adding a `__VLA_\d+D_` case, would close both.

vscode-extension/server/src/completion.ts
```diff
@@ -251,30 +253,65 @@ interface ResolvedTypeInfo {
...
 
+/** One link of a dotted chain: an identifier plus however many subscripts follow it. */
+interface ChainSegment {
+  name: string;
+  subscripts: number;
+}
+
+/** Split "a.b[0].c" into its links, counting the subscripts on each. */
+function parseChainSegments(prefixExpr: string): ChainSegment[] {
+  return prefixExpr.split(".").map((raw) => {
```
thiagoralves: `prefixExpr.split(".")` splits on every dot, including dots *inside* a subscript, so an index expression containing a member access shreds the chain:

```st
VAR arr : ARRAY [0..9] OF Point; s : Idx; END_VAR
arr[s.k].   →  []
```

`getCursorContext` now (correctly) captures `arr[s.k]` as the prefix, but `split(".")` yields `["arr[s", "k]"]`; `"arr[s"` has no balanced bracket so `name` stays `"arr[s"` with `subscripts: 0`, and `scope.lookup("arr[s")` misses. `r := arr[s.k].x;` compiles without error, so this is a legal shape that gets no completions. Splitting only on dots at bracket depth 0 (a small scan instead of `split`) would handle it — same scan would also let `arr[idx[i]].` become a dot-access context, which the `[^\]]*` in the `cursor-context` regex currently rejects outright.

thiagoralves (commented):
Code review — 3 findings on the RTOP-248 half. The RTOP-247 fix (`isStructField` guard) looks correct and complete to me: `StructDefinition` only ever appears at `CompilationUnit.types[].definition`, so the scan covers every struct in the unit, and all five consumers (`hover`, `definition`, `references`, `rename`, `IsWrappedTypeRequest`) already handle a `symbol`-less resolution without re-deriving it from a name. Verified 277/277 on the branch.

vscode-extension/server/src/cursor-context.ts
```diff
@@ -60,7 +60,11 @@ export function getCursorContext(
   const prefix = currentLine.substring(0, column - 1);
 
   // Check dot-access: prefix ends with identifier chain + "."
-  const dotMatch = prefix.match(/([\w]+(?:\.[\w]+)*)\.\s*$/);
+  // Each link may carry subscripts (`pts[i].`, `grid[i][j].`); the chain
```
thiagoralves: The `grid[i][j].` example is misleading: chained subscripts on a *multi-dimensional* array are not valid in this dialect, and this line will send a future reader looking for a bug that isn't one.

```st
VAR grid : ARRAY [0..3, 0..3] OF Point; END_VAR
r := grid[i][i].x;   →  "'GRID' has 2 dimensions but is indexed with 1 index."
```

The compiler's own walk (`type-checker.ts:439`) peels exactly one array level per `[...]` *group*, which is what `unwrapSubscripts` matches — so `grid[i][j].` correctly yields nothing, and the comma form `grid[i, j].` (what the added test actually covers) is the supported one. Chained brackets only mean two levels for a genuine array-of-arrays (`mat : ARRAY [0..3] OF Row`), which does work. Suggest swapping the example to `grid[i, j].` and/or `mat[i][j].`.

vscode-extension/server/src/completion.ts
JoaoGSP: Both confirmed and fixed in `cbadf4a`, in the compiler core rather than worked around here — the gap is in the shared helper and the type-checker's own chain walk calls the same function.

`resolveArrayElementType` now walks `TypeReference` aliases recursively, bounded by the module's existing `MAX_TYPE_ALIAS_DEPTH`, mirroring `resolveArrayShapeByName` directly below it. And it matches `__VLA_<rank>D_<Element>` alongside `__INLINE_ARRAY_`.

Measured on the branch, and your framing was if anything generous to the first pass — the empty list is a regression, not just an unhandled shape:

| shape | before the PR | after the first pass | now |
| --- | --- | --- | --- |
| `ali[i].` alias | 150 items | `[]` | `x, y` |
| `deep[i].` two-hop alias | 150 items | `[]` | `x, y` |
| `vals[i].` `ARRAY [*]` | 150 items | `[]` | `x, y` |

Before, every subscript shape got the generic body list — useless but uniform. After the first pass the shapes the resolver understood became correct and the ones it did not became silently empty, which reads as a broken LSP rather than an unsupported construct. Worth the fix on that ground alone.

Tests added for the alias (including a two-hop chain) and for the VLA in a `VAR_IN_OUT`, both red-green verified. Full compiler suite 2277 passed / 93 files — the core change leaves the type-checker undisturbed.

vscode-extension/server/src/completion.ts
JoaoGSP: Right, and it was mine — introduced by this PR. Fixed in `cbadf4a`.

`parseChainSegments` no longer uses `split(".")`. It scans the prefix tracking bracket depth: dots split only at depth 0, characters inside brackets are skipped entirely, and each top-level `[` group counts as one subscript. So `arr[s.k]` parses as a single link named `arr` with one subscript instead of `["arr[s", "k]"]`.

Took the second half of your suggestion too — the `cursor-context` regex now admits one level of nested brackets, so `arr[idx[i]].` becomes a dot-access context and resolves. Both are covered by tests.

| shape | before the PR | after the first pass | now |
| --- | --- | --- | --- |
| `arr[s.k].` | 150 items | `[]` | `x, y` |
| `arr[idx[i]].` | 150 items | `[]` | `x, y` |

One thing I left alone: arbitrary bracket nesting. The scan handles it, but the regex that decides whether we are in a dot-access context at all only admits one level, so `arr[a[b[i]]].` still falls through to body completions. A full balanced-bracket scan in `getCursorContext` would close that, and it is a rewrite of that function rather than a regex tweak — happy to do it here if you would rather not leave the asymmetry, but I did not think the shape earned it.

vscode-extension/server/src/cursor-context.ts
JoaoGSP: Correct, and the comment is fixed in `cbadf4a`. It now reads:

```
// A link may carry subscripts (`pts[i].`, `grid[i, j].`, `mat[i][j].`), whose
// index may itself be subscripted (`arr[idx[i]].`). The chain resolver unwraps
// one array level per bracket group.
```

Verified your point independently rather than just taking the edit: `mat[i][i].` on a genuine array-of-arrays (`mat : ARRAY [0..3] OF Row`) returns `x, y` both before and after this round — it was never broken — while `grid[i][i]` on the 2-D array is the compile error you quoted. So one level per bracket *group* is exactly the rule, and my example named the one spelling that cannot work.

Added `mat[i][i].` as a test so the array-of-arrays case is pinned rather than incidental, and kept the existing `grid[i, i].` coverage for the comma form.

Thanks for catching this one in particular — a wrong comment costs more than a wrong line, since it survives the next person's reading of correct code.

thiagoralves (approved):

PR #232, opened 2026-09-04
fix(runtime): return the IEC value type from GlobalVar::read() (DOPE-613)

thiagoralves:
Fixes [DOPE-613](https://autonomylogic.atlassian.net/browse/DOPE-613). Reported on the forum against 4.2.1.0 ([a-few-defects-of-4210](https://edge.autonomylogic.com/forum/thread/a-few-defects-of-4210), item 3), with a reproducer project attached.

## The bug

An LD/FBD block taking one **global** and one **local** on its inputs does not compile:

```
pou_MAIN.cpp:15:52: error: no matching function for call to 'ADD(int, strucpp::IEC_INT&)'
    _TMP_ADD593761_OUT = ADD(GI16->read(), LI16);
note: deduced conflicting types for parameter 'T' ('int' and 'strucpp::IECVar<int>')
```

`GlobalVar::read()` unwrapped **two** layers — `GlobalVar` → `IECVar` → raw scalar — while its own doc comment promised "the real IEC value type". The real IEC value type is `IECVar<T>`; `T` is the underlying storage type.

Scalar globals were the only operand kind doing this:

| operand | emitted as | deduces to |
|---|---|---|
| local | `LI16` | `IECVar<int>` |
| literal | `static_cast<IEC_INT>(5)` | `IECVar<int>` |
| GVL / composite global | `GVL->with_lock(…{ return (*__glk).GI16; })` | `IECVar<int>` |
| **scalar global** | `GI16->read()` | **`int`** |

The std-lib templates that share one parameter across both operands — `template<typename T> T ADD(T, T)` — then cannot deduce. `operator T()` cannot rescue it: implicit conversions are not considered during template argument deduction.

**Affected:** `ADD SUB MUL DIV MOD`, `AND OR XOR`, `ATAN2`.
**Affected, also:** `MUX` (a global anywhere among its inputs) and the extensible arity of the arithmetic family, e.g. `ADD(g, l, l)`. An earlier revision of this description listed MUX as unaffected — that was wrong; corrected here and in DOPE-613.

**Not affected:** `GT LT GE LE EQ NE` (already heterogeneous), `MAX MIN LIMIT SEL` (mixed-type overloads added for OSCAT), `SHL`, `EXPT`, GVL globals, and `ADD_TIME` / `SUB_TIME`. LD and FBD always emit the call form, so the graphical languages took the full weight.

**Located variables were affected too** — so "read an input word, add a local offset" was broken.

### One deliberate behaviour change

Infix ST on scalar globals is **not** untouched, contrary to an earlier revision of this description. A scalar global now presents as `IECVar<T>`, so infix binds the `IECVar` operator overloads, which truncate to the IEC width at each step instead of promoting to `int`. For `ga`, `gb` SINT = 100, `o := (ga + gb) / 2` was `100` and is now `-28`.

That is the correct answer, not a regression: two SINT *locals* already gave `-28`, and IEC 61131-3 says an operation on SINT yields SINT — so globals and locals now agree. It was an unpinned change, so a test now asserts that agreement.

The reporter's second claim, that "WORD is not supported by ADD", is a misattribution of this same bug: `ADD(Lw1, Lw2)` on two local WORDs builds and links fine. Their WORD operand was a global.

## The fix

Return `V`. That restores the invariant the rest of the library already assumes, so no std-lib signature changes are needed and every affected function is fixed at once. The return type is spelled `V` rather than `auto` so it cannot drift again.

**Rejected alternative:** heterogeneous `(A a, B b)` signatures across the arithmetic and bit families. Beyond being ~10 functions × 2 arities, it changes the return type — `T ADD(T,T)` returns `T(iec_unwrap(a) + iec_unwrap(b))`, and that explicit `T(...)` truncates back to the IEC width. An `auto` return would let `SINT + SINT` promote to `int` and stop wrapping at 8 bits, silently changing overflow behaviour of code that works today.

## What is unchanged

- **Forcing.** `IECVar`'s copy constructor is deliberately forcing-collapsing (`value_{other.get()}, forced_{false}`), so the snapshot carries the effective forced value with the flag cleared — what a detached read should return. The two STRING/WSTRING wrappers had `= default` special members and so did **not** have those semantics; returning `V` exposed that, and the second commit gives them IECVar's behaviour. See the review thread — this was a genuine blocker, caught by review rather than by the suite.
- **Threading.** Still one lock acquisition; the copy is constructed while the `lock_guard` is in scope, so the snapshot stays atomic.
- **Size — for scalars.** Byte-identical on ATmega2560 at `-Os` for INT / LINT / REAL globals: `run()` frame 3 → 3, `.text` unchanged. The wrapper copy is fully elided.
- **Size — strings cost stack.** A STRING global is a 518-byte wrapper, and returning it by value is not elided. For a body of `o := gs;`, `run()`'s frame on ATmega2560 at `-Os` goes 262 → 520 bytes. An earlier revision of this description claimed "identical object size" without qualification; that measurement was taken on the scalar case only. Called out rather than avoided: keeping `read()` uniform is the point of the fix, and a per-type exception is exactly the asymmetry that caused DOPE-613 in the first place.

## Testing

`tests/integration/global-local-operand-mix-cpp.test.ts` — 34 tests, end-to-end through g++ on purpose: the defect lived entirely in C++ overload resolution, so a codegen-string assertion would have stayed green throughout.

**16 of the 34 fail without this change.** The 18 that pass regardless are the matched-operand controls (all-global, all-local), kept deliberately so a future regression is attributed to the global path rather than to the function itself.

Coverage:
- the mixed-operand matrix across every affected function
- all-global / all-local controls
- the value actually reaching the block (`7 + 3 = 10`, not a default-constructed operand)
- forcing through a block input, and that a write to a forced global is still ignored
- that the returned snapshot is detached — mutating it must not reach the canonical
- three real-threaded tests under `-DSTRUCPP_THREADED -pthread`: 4 writers hammering a 64-bit global while readers observe through `read()`, asserting zero torn values, plus a concurrent force/unforce case. Each asserts the reader actually progressed, so a deadlock cannot pass as a clean run.

Full suite green: **2356 passed**, 94 files. Typecheck clean, lint clean (`src/`), Prettier clean.

## Hardware verification

Built a real `openplc-cli` from this branch (packed strucpp, installed it the way `download-binaries.ts` does, rebuilt the CLI bundle) rather than patching generated output.

- **The reporter's own project** now compiles for ATmega2560 — same emitted call, `ADD(GI16->read(), LI16)`, now accepted.
- **AutomationDirect P1AM-100** (baremetal, real upload over serial), read back through the Modbus RTU debugger:

  | variable | expression | expected | on device |
  |---|---|---|---|
  | `qAdd` | `ADD(gIn, lBase)` | 10 | 10 |
  | `qSub` | `SUB(gIn, lBase)` | 4 | 4 |
  | `qMul` | `MUL(gIn, lBase)` | 21 | 21 |
  | `qAnd` | `AND(gwIn, lMask)` | `16#000F` | `16#000F` |
  | `qLoc` | `ADD(%IW0, lBase)` | `%IW0 + 3` | 3 |

  Forcing `gIn := 100` on the running PLC moved them to 103 / 97 / 300.
- **OpenPLC Runtime v4** (the `STRUCPP_THREADED` build) compiles from the same project.

Not covered: an upload to the SLM-RP4. It has no user account right now, and creating one is not something I will do unprompted.

## Follow-up (not in this PR)

`openplc-editor` pins strucpp in `binary-versions.json` (currently `v0.6.6`). It needs a bump once this merges and a release is cut — the editor cannot pick the fix up before then.

[DOPE-613]: https://autonomylogic.atlassian.net/browse/DOPE-613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/runtime/include/iec_global.hpp
+++ b/src/runtime/include/iec_global.hpp
@@ -35,10 +35,17 @@
  *     arrays, and FB instances all work, and different-field writes from
  *     different tasks all survive because they mutate the one shared object.
  *
- * read() returns the real IEC value type, so generated bodies and constrained
- * std-lib templates (NOT, ADD, …) deduce the operand correctly — the wrapper is
- * never the deduced operand. Forcing is preserved via the canonical's
- * get()/set(); located globals additionally honor the image forced-slot bitmap.
+ * read() returns the real IEC value type — `V` itself, i.e. `IECVar<T>`, NOT the
+ * underlying `T`. That matters for deduction: the constrained std-lib templates
+ * take one shared parameter for both operands (`template<typename T> T ADD(T, T)`),
+ * and locals, literals and composite-global fields all present as `IECVar<T>`.
+ * Returning `T` here would make a scalar global the one operand kind that differs,
+ * and `ADD(aGlobal, aLocal)` would fail to deduce — implicit conversions are not
+ * considered during template argument deduction, so `operator T()` cannot rescue it.
+ * Forcing is preserved: the copy handed back is built through the canonical's
+ * get(), so it carries the effective forced value (with the force flag cleared, as
+ * a detached snapshot should); located globals additionally honor the image
+ * forced-slot bitmap.
  */
 #ifndef STRUCPP_IEC_GLOBAL_HPP
 #define STRUCPP_IEC_GLOBAL_HPP
@@ -71,12 +78,20 @@ class GlobalVar {
 
     /** Scalar read: returns the real IEC value type (forcing-aware),
      *  deduction-friendly. Only instantiated for scalar globals (codegen uses
-     *  with_lock() for structs / arrays / FB instances). */
-    auto read() const {
+     *  with_lock() for structs / arrays / FB instances).
+     *
+     *  The return type is spelled `V` rather than `auto` on purpose: `auto` let
+     *  this silently return the underlying scalar, which is what broke deduction
+     *  for every mixed global/local operand pair (see the header note above).
+     *
+     *  The copy is constructed while `lg` is still in scope, so the snapshot is
+     *  taken atomically exactly as a `value.get()` would have been — same single
+     *  lock acquisition, same hold duration. */
+    V read() const {
 #ifdef STRUCPP_THREADED
         std::lock_guard<std::mutex> lg(mtx_);
 #endif
-        return value.get();
+        return value;
     }
 
     /** Scalar write (forcing-aware via set()). */
--- a/src/runtime/include/iec_string.hpp
+++ b/src/runtime/include/iec_string.hpp
@@ -351,10 +351,35 @@ public:
     IECStringVar() noexcept : value_{}, forced_{false}, forced_value_{} {}
     IECStringVar(const value_type& v) noexcept : value_{v}, forced_{false}, forced_value_{} {}
     IECStringVar(const char* str) noexcept : value_{str}, forced_{false}, forced_value_{} {}
-    IECStringVar(const IECStringVar&) = default;
-    IECStringVar(IECStringVar&&) = default;
-    IECStringVar& operator=(const IECStringVar&) = default;
-    IECStringVar& operator=(IECStringVar&&) = default;
+    /* Forcing semantics, matching IECVar (see iec_var.hpp).
+     *
+     * These cannot be `= default`. A memberwise copy carries `forced_` and
+     * `forced_value_` across, which breaks forcing two ways: assigning FROM a
+     * forced source leaks the flag onto the destination (whose next `set()` is
+     * then silently dropped), and assigning from an unforced source clears a
+     * force the debugger is holding. Generated code assigns every scan, so
+     * either one destroys a force within one cycle -- and STRING forcing is a
+     * first-class debug operation (force_string / unforce_string).
+     *
+     * Construction takes a detached snapshot of the effective value and starts
+     * unforced; assignment routes through set() so the destination keeps its
+     * own forcing state. */
+    IECStringVar(const IECStringVar& other) noexcept
+        : value_{other.forced_ ? other.forced_value_ : other.value_},
+          forced_{false},
+          forced_value_{} {}
+    IECStringVar(IECStringVar&& other) noexcept
+        : value_{other.forced_ ? other.forced_value_ : other.value_},
+          forced_{false},
+          forced_value_{} {}
+    IECStringVar& operator=(const IECStringVar& other) noexcept {
+        set(other.forced_ ? other.forced_value_ : other.value_);
+        return *this;
+    }
+    IECStringVar& operator=(IECStringVar&& other) noexcept {
+        set(other.forced_ ? other.forced_value_ : other.value_);
+        return *this;
+    }
 
     // Cross-size converting constructor (IEC 61131-3: STRING types are interoperable)
     template<size_t OtherLen, std::enable_if_t<OtherLen != MaxLen, int> = 0>
--- a/src/runtime/include/iec_wstring.hpp
+++ b/src/runtime/include/iec_wstring.hpp
@@ -369,10 +369,35 @@ public:
     IECWStringVar() noexcept : value_{}, forced_{false}, forced_value_{} {}
     IECWStringVar(const value_type& v) noexcept : value_{v}, forced_{false}, forced_value_{} {}
     IECWStringVar(const char16_t* str) noexcept : value_{str}, forced_{false}, forced_value_{} {}
-    IECWStringVar(const IECWStringVar&) = default;
-    IECWStringVar(IECWStringVar&&) = default;
-    IECWStringVar& operator=(const IECWStringVar&) = default;
-    IECWStringVar& operator=(IECWStringVar&&) = default;
+    /* Forcing semantics, matching IECVar (see iec_var.hpp).
+     *
+     * These cannot be `= default`. A memberwise copy carries `forced_` and
+     * `forced_value_` across, which breaks forcing two ways: assigning FROM a
+     * forced source leaks the flag onto the destination (whose next `set()` is
+     * then silently dropped), and assigning from an unforced source clears a
+     * force the debugger is holding. Generated code assigns every scan, so
+     * either one destroys a force within one cycle -- and STRING forcing is a
+     * first-class debug operation (force_string / unforce_string).
+     *
+     * Construction takes a detached snapshot of the effective value and starts
+     * unforced; assignment routes through set() so the destination keeps its
+     * own forcing state. */
+    IECWStringVar(const IECWStringVar& other) noexcept
+        : value_{other.forced_ ? other.forced_value_ : other.value_},
+          forced_{false},
+          forced_value_{} {}
+    IECWStringVar(IECWStringVar&& other) noexcept
+        : value_{other.forced_ ? other.forced_value_ : other.value_},
+          forced_{false},
+          forced_value_{} {}
+    IECWStringVar& operator=(const IECWStringVar& other) noexcept {
+        set(other.forced_ ? other.forced_value_ : other.value_);
+        return *this;
+    }
+    IECWStringVar& operator=(IECWStringVar&& other) noexcept {
+        set(other.forced_ ? other.forced_value_ : other.value_);
+        return *this;
+    }
 
     // Cross-size assignment (IEC 61131-3: WSTRING types are interoperable, truncation on overflow)
     template<size_t OtherLen>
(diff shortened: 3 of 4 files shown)


marconetsf:
Reviewed with a local validation pass. The full suite is green on this branch (**2356 passed / 94 files**, matching the PR body), and the negative control reproduces exactly — reverting `read()` to `value.get()` gives **16 failed / 18 passed of 34**. The root-cause analysis holds and the fix is appropriately minimal. Four things came out of the pass, one of which I think blocks.

## 1. Blocker — the change silently breaks forcing on STRING/WSTRING globals

`read()` now returns `V`, which for a STRING global is `IECStringVar<254>`. That wrapper's copy constructor and copy assignment are `= default`, i.e. memberwise (`src/runtime/include/iec_string.hpp:354-357`) — unlike `IECVar`, whose copy ctor deliberately collapses forcing (`iec_var.hpp:93`) and whose copy assignment routes through `set()` (`iec_var.hpp:110`) precisely so that generated code doing `BLINK := TOF0.Q` every scan cannot unforce a variable the debugger is holding. `src/runtime/include/iec_wstring.hpp:372-375` has the same shape.

So `O = GS->read()` now memberwise-copies force state. Two observable failures, both measured on `development` vs. this branch from the same generated C++, for the ST body `o := gs;` (STRING global into a local STRING):

| scenario | development | this branch |
|---|---|---|
| local `o` forced to `'FF'`, global unforced | `o='FF'`, forced=1 | `o='AB'`, **forced=0** — the engineer's force is cleared every scan |
| global forced to `'ZZ'`, then `o := 'QQ';` | `o='QQ'` | `o='ZZ'`, forced=1 — the flag leaks onto the local and the next write is silently dropped |

This is reachable from the product: STRING forcing is a first-class debug operation (`force_string` / `unforce_string`, `src/runtime/include/debug_dispatch.hpp:180-187`).

Suggested fix — give the two string wrappers the forcing semantics `IECVar` already has:

```cpp
IECStringVar(const IECStringVar& other) noexcept : value_{other.get()}, forced_{false}, forced_value_{} {}
IECStringVar(IECStringVar&& other) noexcept : value_{other.get()}, forced_{false}, forced_value_{} {}
IECStringVar& operator=(const IECStringVar& other) noexcept { set(other.get()); return *this; }
IECStringVar& operator=(IECStringVar&& other) noexcept { set(other.get()); return *this; }
```

I applied that to both headers locally: it fixes both rows above and the full suite stays green (2356 passed / 94 files). It also makes the new header note at `src/runtime/include/iec_global.hpp:45-48` ("with the force flag cleared, as a detached snapshot should") true for strings, which today it is not.

## 2. The size claim does not hold for STRING globals

`sizeof(IEC_STRING)` is 518 against 258 for `IECString<254>`, and the copy is not elided. For a program whose whole body is `o := gs;`, `Program_MAIN::run()` at `-Os` (`-fstack-usage`, x86-64):

- `development`: 288 bytes
- this branch: 544 bytes

"Identical object size on ATmega2560" was measured on the scalar case; +256 bytes of frame per string-global read is not noise on 8 KB of SRAM. Worth either scoping the claim to scalars or measuring it on the AVR target.

## 3. Infix ST on scalar globals is no longer unaffected

The PR body lists infix ST as untouched because it "lowers to native C++ operators". After this change a scalar global presents as `IECVar<T>`, so infix on globals binds the `IECVar` operator overloads, which truncate to the IEC width at every step instead of promoting to `int`:

```
VAR_GLOBAL ga : SINT := 100; gb : SINT := 100; END_VAR
o := (ga + gb) / 2;
```

`development`: `100`. This branch: `-28`.

I don't think this should be reverted — two SINT *locals* already give `-28`, and IEC 61131-3 says an operation on SINT yields SINT, so this makes globals consistent. But it is a silent change of overflow behaviour shipping in a bug fix, and nothing pins it. Worth a line in the PR body and a test.

## 4. MUX was affected, and so was the variadic call form

"Not affected: ... `MAX MIN LIMIT SEL MUX`" does not hold for MUX. Compiling the generated C++ against each header (`-fsyntax-only`, error count):

| ST | development | this branch |
|---|---|---|
| `o := MUX(gi, l, gi, l);` | 1 error | 0 |
| `o := ADD(gi, l, l);` | 1 error | 0 |
| `MAX` / `MIN` / `LIMIT` / `SEL` / `SHL` / `EXPT` / `GT` / `EQ` with `(gi, l)` | 0 | 0 |

So the list is right except for MUX, and the 3-operand call form of the arithmetic family broke as well. Neither is in the `AFFECTED` matrix at `tests/integration/global-local-operand-mix-cpp.test.ts:83-99`. The same MUX claim is in the DOPE-613 description, so that needs the same correction.

Related coverage gap worth closing while you are in there: the matrix has no STRING/WSTRING global, no located global (`AT %IW0`, which did break) and no TIME (`ADD(gtm, lt)` also broke). The STRING gap is why the blocker above got past 2356 green tests.

thiagoralves:
Thanks — this was a genuinely good pass. I verified all four independently rather than taking them on trust; **all four hold**, and one of the closing side-claims does not. Both fixes are pushed (`4689d02`).

## 1. Blocker — confirmed and fixed

Reproduced exactly, same generated C++, swapping only the header: `A o=AB forced=0` / `B o=ZZ forced=1 → set('QQ') stays ZZ`, against `A o=FF forced=1` / `B o=ZZ forced=0 → QQ` on development. Your suggested shape is what landed, with one change: the copies reference the effective member directly rather than going through `get()`. `get()` returns `value_type` **by value**, so the literal form cost an extra 258-byte temporary and pushed `run()` on ATmega2560 from 520 to **777** bytes. Referencing the member keeps it at 520 — so the forcing fix is free.

## 2. Size claim — confirmed, and worse than your x86 figure in context

You measured 288→544 on x86-64. On the AVR target my claim actually named, `run()` frame at `-Os`:

| global type | development | this branch |
|---|---|---|
| INT / LINT / REAL | 3 | 3 (`.text` identical) |
| **STRING** | **262** | **520** |

Same +258. PR body and DOPE-613 both now scope the claim to scalars and state the string figure. I kept `read()` uniform rather than special-casing strings: a per-type exception to the operand representation is precisely the asymmetry that caused this bug, and a future ANY_STRING template with a shared parameter would resurrect it silently.

## 3. Infix overflow — confirmed, now pinned

SINT globals 100/100, `(ga + gb) / 2`: development `100`, branch `-28`. Agreed it should not be reverted — two SINT locals already gave `-28` and the standard says SINT op SINT yields SINT. A test now asserts globals and locals agree, so it stops being unpinned, and the PR body calls it out as a deliberate change.

## 4. MUX — you are right, I was wrong

`MUX(g, l, g, l)` and `ADD(g, l, l)` both: 1 error on development, 0 here. Corrected in the PR body and in DOPE-613, and both are in the matrix now, along with a located global (`AT %IW0`).

## Closing note: TIME does not hold

The other two do. But `ADD(gtm, lt)` never reaches C++ — the ST type checker rejects it first with *"Argument 'IN1' of 'ADD' expects ANY_NUM, got TIME"*. The correct forms, `ADD_TIME` and `SUB_TIME`, report **0 errors on development** — never broken. So no TIME coverage was added; happy to be shown a shape I missed.

## Discrimination and totals

Reverting only the string-wrapper change fails exactly the 2 string tests; reverting only `iec_global.hpp` fails 20 of 40. Full suite **2362 passed / 94 files**.

## An unrelated bug your blocker led me to

Trying to verify the string fix on hardware, every STRING force was refused with `ERROR_OUT_OF_MEMORY`. That is two separate bugs, neither related to this PR:

1. **A length-contract mismatch that breaks STRING forcing on every target.** `handle_set` (`debug_dispatch.hpp:383`) requires `len >= type_ops[tag].size`, which for STRING is `DEBUG_STRING_WIDTH` = 127. The editor/CLI encoder (`encodeForceValue`, `len8-utf8`) sends a compact `1 + text.length` payload — 3 bytes for `"'FF'"`. 3 < 127, so it is rejected regardless of value. Scalars agree on both sides, which is why numeric forcing works and this went unnoticed.
2. **Status `0x82` is double-booked.** Device `STATUS_DATA_TOO_LARGE = 0x82`; client `ERROR_OUT_OF_MEMORY = 0x82`. So a length mismatch reports as exhausted memory — and `useDebugPolling` responds by halving its batch size, which is right for a real OOM on the read path and useless here.

Proven rather than inferred: padding the encoder to the full 127-byte width makes it work on a P1AM-100 first try — `gStr = ZZ [FORCED]`. I reverted that diagnostic; the real fix is a decision between padding client-side (works with firmware already in the field) and relaxing the device gate to `len >= 1 + bytes[0]` for the string tags (cleaner, needs a reflash). WSTRING forcing is separately unimplemented client-side (`len8-utf16le` throws "not supported yet").

That also let me verify item 1 on real hardware after all: with the padded payload, forcing the local `oStr` to `'FF' holds as `FF [FORCED]` across a 20 ms scan that assigns `oStr := gStr` every cycle, and unforcing hands it back to the program. Before the string-wrapper fix that force would have been cleared within one cycle.

marconetsf (approved):

PR #233, opened 2026-09-04
fix(debug): validate a string payload by its length prefix (DOPE-614)

thiagoralves:
Fixes [DOPE-614](https://autonomylogic.atlassian.net/browse/DOPE-614). Found while verifying DOPE-613 on hardware.

## The bug

Forcing a STRING is refused on **every** target, reported as exhausted memory:

```
$ openplc-cli debug force Config0:gStr ZZ
error [target_error]: ERROR_OUT_OF_MEMORY
```

A string is length-prefixed on the wire — `bytes[0]` is the character (STRING) or code-unit (WSTRING) count, and `force_string` / `write_string` read exactly that many. But the gate required `len >= type_ops[tag].size`, and for the string tags that size is the **padded field width the read path emits**: `DEBUG_STRING_WIDTH` 127, `DEBUG_WSTRING_WIDTH` 253.

Every caller sends a compact payload. The editor encodes `1 + text.length`, so `"ZZ"` is 3 bytes — and `3 < 127` was refused regardless of value or target. Scalars agree on both sides (`expected == sizeof(T)`, and the encoder emits exactly that), which is why only strings broke and why this went unnoticed.

This is not baremetal-specific. On Runtime v4 the transport is fine — 4096-byte frame, 256-byte journal payload — the force is enqueued and then applied through this same `handle_set`, and refused there.

## The fix

Validate the prefix instead of the padded width. Safe both directions: too short is still refused, so a prefix that overruns its own payload cannot make the read walk off the end; and `len` stays a lower bound, so a caller that pads to the full width keeps working.

**`handle_write` had the identical bug**, so soft writes to a string — the OPC-UA write path and the retain restore walk — were refused too. The rule now lives in one shared `validate_payload()` used by both, because it was duplicated inline and wrong in both copies.

## Scope — deliberately just the set path

The read path stays fixed-width, so **no client change is needed and nothing releases in lockstep**: an existing client works against a new device immediately.

Variable-length *reads* would additionally unbreak STRING debug on the 128-byte-frame AVRs (ATmega328P/168/32U4/16U4, where a 127-byte field already never fits) and WSTRING on baremetal. That changes the wire format for three independent readers — this dispatch, the baremetal Modbus slave in `openplc-editor/resources/sources/Baremetal/`, and `debug_handler.c` in `openplc-runtime` — plus the client stride and both web mirrors. Left out on purpose; recorded on the ticket.

Also not fixed here: **status `0x82` is triple-booked** — `STATUS_DATA_TOO_LARGE` here, `ERROR_OUT_OF_MEMORY` in the client, and a full write journal in `openplc-runtime`. That mislabel is why this looked like a device memory limit rather than a length-contract bug. Tracked separately.

## Testing

`tests/integration/dope614-string-force-length.test.ts` drives the dispatch entry points directly — the gate is the whole defect, and both transports hand it the same bytes.

**5 of the 7 fail on `development`.** The 2 that pass regardless are the padded-payload guard and the scalar guard — kept so a future regression is attributed to the string rule rather than to the gate disappearing.

(An earlier revision of this description named the wrong two. The "still refuses a payload shorter than its own prefix claims" test *fails* on `development`, for the reason in the next section.)

Covered: compact STRING; compact WSTRING (code units, not bytes, so `1 + 2n` — getting that wrong would accept a truncated payload and read past it); no stale tail when a short force follows a long one; a prefix that overruns its payload; a count past the cap; a null payload; scalars unchanged; and a force/unforce round trip.

That last one earned its keep: it asserts a write while forced returns `STATUS_OK` and is *then* dropped by `IECVar::set()`. Before `handle_write` was fixed the value assertion passed for the wrong reason — the write was being refused outright, not ignored because of the force.

Full suite: **2370 passed / 95 files**. Typecheck, lint and Prettier clean.

## One behaviour change worth naming

An over-cap prefix is now refused instead of silently truncated. On `development`, `force_string` clamps `wire_len` to `DEBUG_STRING_CAP`, so a payload declaring `bytes[0] = 200` with 200 bytes behind it passed the old `len >= 127` check and was accepted — the value came back mangled. `validate_payload` now returns `STATUS_DATA_TOO_LARGE` for any count above the cap.

That is the reason the "still refuses" test fails on `development` rather than passing as a control.

Nothing of ours sends such a payload: the editor's encoder throws above 126 (`variable-sizes.ts:431`), and the CLI shares that encoder. Only a third-party client emitting a malformed prefix would notice, and refusing it is the correct answer — truncating an engineer's force to a value they did not ask for is worse than rejecting it.

## Hardware

P1AM-100, CLI rebuilt from this branch with **the client encoder untouched**:

| action | result |
|---|---|
| force `gStr` = `ZZ` (3-byte payload) | `ZZ [FORCED]` |
| force `gStr` = `HELLOWORLD1` (12 B) | `HELLOWORLD1 [FORCED]` |
| force `gStr` = `AB` again (3 B) | `AB [FORCED]` — no stale characters |
| force STRING local `oStr` = `FF` | holds `FF [FORCED]` across a 20 ms scan that assigns over it every cycle |
| numerics | unchanged at 10 / 4 / 21 / `16#000F` |

[DOPE-614]: https://autonomylogic.atlassian.net/browse/DOPE-614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/runtime/include/debug_dispatch.hpp
+++ b/src/runtime/include/debug_dispatch.hpp
@@ -365,6 +365,47 @@ inline Entry read_entry(uint8_t arr, uint16_t elem) noexcept {
 // Per-entry operations. These are what ModbusSlave / Runtime v4 call.
 // ---------------------------------------------------------------------------
 
+/**
+ * Validate a value payload against a leaf's type. Returns STATUS_OK, or the
+ * STATUS_* refusal to hand straight back to the caller.
+ *
+ * Shared by handle_set() and handle_write() deliberately: they had the same
+ * rule inline and it was wrong in both, so a single definition is what keeps
+ * them from drifting apart again.
+ *
+ * Scalars are fixed-width, so `len` must cover the type's size.
+ *
+ * Strings are LENGTH-PREFIXED on the wire: `bytes[0]` is the character
+ * (STRING) or code-unit (WSTRING) count, and force_string / write_string read
+ * exactly that many. `type_ops[tag].size` is the PADDED field width the READ
+ * path emits (DEBUG_STRING_WIDTH / DEBUG_WSTRING_WIDTH), so requiring it here
+ * refused every compact payload a caller actually sends -- which is what broke
+ * STRING forcing and soft-writes on every target (DOPE-614).
+ *
+ * Validating the prefix instead is safe in both directions. Too short is still
+ * refused, so a prefix that overruns its own payload cannot make the read walk
+ * off the end. And `len` stays a lower bound, so a caller that pads to the full
+ * width keeps working unchanged.
+ */
+inline uint8_t validate_payload(uint8_t tag, const uint8_t* bytes, uint16_t len) noexcept {
+    const uint8_t expected = type_ops[tag].size;
+    if (expected == 0) return STATUS_DATA_TOO_LARGE;
+    if (!bytes) return STATUS_DATA_TOO_LARGE;
+
+    if (tag == TAG_STRING || tag == TAG_WSTRING) {
+        const uint8_t count = bytes[0];
+        if (count > DEBUG_STRING_CAP) return STATUS_DATA_TOO_LARGE;
+        const uint16_t need = static_cast<uint16_t>(
+            1u + (tag == TAG_WSTRING ? static_cast<uint16_t>(count) * 2u
+                                     : static_cast<uint16_t>(count)));
+        if (len < need) return STATUS_DATA_TOO_LARGE;
+        return STATUS_OK;
+    }
+
+    if (len < expected) return STATUS_DATA_TOO_LARGE;
+    return STATUS_OK;
+}
+
 /** Set (force or unforce) a variable. Returns STATUS_* code. */
 inline uint8_t handle_set(uint8_t arr, uint16_t elem, bool forcing,
                           const uint8_t* bytes, uint16_t len) noexcept {
@@ -377,10 +418,8 @@ inline uint8_t handle_set(uint8_t arr, uint16_t elem, bool forcing,
     if (e.flags & LEAF_FLAG_READONLY) return STATUS_READ_ONLY;
 
     if (forcing) {
-        uint8_t expected = type_ops[e.tag].size;
-        // size == 0 is the string stub — Phase 4a rejects for now
-        if (expected == 0) return STATUS_DATA_TOO_LARGE;
-        if (len < expected) return STATUS_DATA_TOO_LARGE;
+        const uint8_t bad = validate_payload(e.tag, bytes, len);
+        if (bad != STATUS_OK) return bad;
         type_ops[e.tag].force(e.ptr, bytes);
     } else {
         type_ops[e.tag].unforce(e.ptr);
@@ -413,9 +452,8 @@ inline uint8_t handle_write(uint8_t arr, uint16_t elem,
     // uses, so a CONSTANT can never be clobbered by a stale retained value
     // either — constants come from the declaration, never from storage.
     if (e.flags & LEAF_FLAG_READONLY) return STATUS_READ_ONLY;
-    uint8_t expected = type_ops[e.tag].size;
-    if (expected == 0) return STATUS_DATA_TOO_LARGE;  // string stub
-    if (len < expected) return STATUS_DATA_TOO_LARGE;
+    const uint8_t bad = validate_payload(e.tag, bytes, len);
+    if (bad != STATUS_OK) return bad;
     type_ops[e.tag].write(e.ptr, bytes);
     return STATUS_OK;
 }
--- a/src/runtime/include/debug_dispatch.hpp
+++ b/src/runtime/include/debug_dispatch.hpp
@@ -367,25 +367,19 @@ inline Entry read_entry(uint8_t arr, uint16_t elem) noexcept {
 
 /**
  * Validate a value payload against a leaf's type. Returns STATUS_OK, or the
- * STATUS_* refusal to hand straight back to the caller.
+ * STATUS_* refusal to hand straight back to the caller. Shared by handle_set()
+ * and handle_write() so the rule cannot drift between them.
  *
- * Shared by handle_set() and handle_write() deliberately: they had the same
- * rule inline and it was wrong in both, so a single definition is what keeps
- * them from drifting apart again.
+ * Scalars are fixed-width: `len` must cover the type's size.
  *
- * Scalars are fixed-width, so `len` must cover the type's size.
+ * Strings are length-prefixed: `bytes[0]` is the character (STRING) or
+ * code-unit (WSTRING) count, and force_string / write_string read exactly that
+ * many, so `len` must cover `1 + count` -- `1 + 2 * count` for WSTRING -- and
+ * NOT `type_ops[tag].size`, which is the padded width the READ path emits. A
+ * count past DEBUG_STRING_CAP is refused rather than silently truncated.
  *
- * Strings are LENGTH-PREFIXED on the wire: `bytes[0]` is the character
- * (STRING) or code-unit (WSTRING) count, and force_string / write_string read
- * exactly that many. `type_ops[tag].size` is the PADDED field width the READ
- * path emits (DEBUG_STRING_WIDTH / DEBUG_WSTRING_WIDTH), so requiring it here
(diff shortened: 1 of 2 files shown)


src/runtime/include/debug_dispatch.hpp
```diff
@@ -365,6 +365,47 @@ inline Entry read_entry(uint8_t arr, uint16_t elem) noexcept {
...
+ * width keeps working unchanged.
+ */
+inline uint8_t validate_payload(uint8_t tag, const uint8_t* bytes, uint16_t len) noexcept {
+    const uint8_t expected = type_ops[tag].size;
+    if (expected == 0) return STATUS_DATA_TOO_LARGE;
+    if (!bytes) return STATUS_DATA_TOO_LARGE;
+
+    if (tag == TAG_STRING || tag == TAG_WSTRING) {
+        const uint8_t count = bytes[0];
+        if (count > DEBUG_STRING_CAP) return STATUS_DATA_TOO_LARGE;
```
JoaoGSP, line 391: Nit: worth one line in the PR body. On `development` a prefix above 126 inside a padded payload was accepted and silently truncated (the negative control shows the value ending up empty, `over cap: got 0x7E want 0x82`); this line now refuses it with 0x82. Correct, but it is a behaviour change for any third-party client. The editor and CLI never send it (the encoder throws at `variable-sizes.ts:431`), so nothing of ours is affected.

src/runtime/include/debug_dispatch.hpp
```diff
@@ -365,6 +365,47 @@ inline Entry read_entry(uint8_t arr, uint16_t elem) noexcept {
 // Per-entry operations. These are what ModbusSlave / Runtime v4 call.
 // ---------------------------------------------------------------------------
 
+/**
```
JoaoGSP, line 368: Nit: this 20-line comment carries the ticket key and the history the commit message and PR already hold ("which is what broke STRING forcing ... (DOPE-614)"). Keep the contract the reader needs at this spot, the prefix rule and that `len` is a lower bound, and drop the narrative.

JoaoGSP (commented):
Reviewed the diff and ran the full local gate in a throwaway worktree (CI runs on `main` only, so a PR to `development` gets no checks). The gate is correct and the fix is proven; what is left is process.

**Verified**
- Bounds: the gate guarantees `count <= DEBUG_STRING_CAP` and `len >= 1 + count` (`x2` for WSTRING); `force_string` / `force_wstring` read at most `count` units from `bytes + 1`; `IECString(ptr, len)` clamps to MaxLen 254; the `char16_t buf[126]` is indexed below `wire_len`. Null payload is now refused. Scalars unchanged. Unforce needs no payload.
- One `validate_payload` for both `handle_set` and `handle_write`; `tag` is bounds-checked by both callers before it indexes `type_ops`.
- Baremetal `debugSetTrace` (`openplc-editor` `modbus_debug.cpp:109`, `len > MAX_MB_FRAME - 8`) is unchanged and a compact payload passes it. The editor encoder already emits `1 + text.length` and caps at 126, so no client change is needed, as stated.
- New test 7/7. Negative control with `development`'s header: 5 failed / 2 passed, exactly the two the PR names. Full suite 2370 passed / 95 files / 7 skipped. Typecheck clean, lint 0 errors, Prettier clean on `src/` and the test file.
- Not verifiable here: the P1AM-100 table and the Runtime v4 journal path.

**Required (process, not code)**
1. AC2 (the 0x82 mislabel) and AC4 (128-byte-frame boards and WSTRING) are deferred as "tracked separately" and "recorded on the ticket", but I could not find a Jira issue for either (searched DOPE and RTOP since 2026-09-01 for 0x82, ERROR_OUT_OF_MEMORY, STATUS_DATA_TOO_LARGE, variable-length, WSTRING debug), and DOPE-614 carries no deferral note beyond the 12:37 approach comment. Please file both follow-ups and link them, or amend the acceptance criteria on the ticket.
2. DOPE-614 has no assignee and sits in In Development with this PR open.
3. Question, not a blocker: this edits the parser of an externally supplied payload on the debug protocol, which is one of the CRA trigger areas, so the bug-fix exemption does not cover it on its own. The change is strictly tightening (null check, `count <= 126`, `len >= 1 + count`), so a short "no new attack surface" note would close it, but someone has to decide and write it.
4. Nit: the process asks a bug-fix PR to state in one line that no requirements document exists and why. The body only says "Fixes DOPE-614".

Fix reaches devices only after a strucpp release plus a `binary-versions.json` bump in openplc-editor and openplc-web, the same release RTOP-247/248 are waiting on.

src/runtime/include/debug_dispatch.hpp
thiagoralves: Good catch, and it was worse than a nit — it exposed an error in the PR body.

Confirmed your reading: on `development`, `force_string` clamps `wire_len` to `DEBUG_STRING_CAP`, so `bytes[0] = 200` with 200 bytes behind it passed the old `len >= 127` check and was accepted with a mangled value. This line refuses it with 0x82.

The knock-on I had missed: that is exactly why the "still refuses a payload shorter than its own prefix claims" test **fails** on `development`. My PR body claimed the two passing controls were that one plus the padded-payload guard. I re-ran the negative control — the two that actually pass are the **padded-payload guard and the scalar guard**. So my description of my own control was wrong, not just imprecise.

Both fixed in the body: the control sentence now names the right two tests, and there is a new section stating the over-cap behaviour change, that nothing of ours emits such a payload (the editor encoder throws above 126 at `variable-sizes.ts:431`, and the CLI shares it), and why refusing beats truncating — a force silently reduced to a value the engineer did not ask for is worse than a rejection.

src/runtime/include/debug_dispatch.hpp
thiagoralves: Agreed — trimmed in `20f81ee`.

The comment now carries only what a reader needs at that spot: scalars are fixed-width; strings are length-prefixed so the requirement is `1 + count` (`1 + 2 * count` for WSTRING) rather than `type_ops[tag].size`, which is the padded width the read path emits; an over-cap count is refused rather than truncated; and `len` is a lower bound throughout. The ticket key and the account of what broke are gone — the commit message and PR body hold that.

thiagoralves:
Thanks — I verified all of it independently rather than taking it on trust. Both inline threads are fixed and resolved; the four process items are answered below, three of them as deliberate declines by the repo owner.

## Fixed

**Over-cap prefix (thread on :397).** Valid, and it exposed an error in the PR body rather than just needing a line. That behaviour change is why the "still refuses" test **fails** on `development` — so my claim about which two tests pass as controls was wrong. The real pair is the padded-payload guard and the scalar guard. Body corrected on both points, with a section naming the change and why refusing beats truncating.

**Comment (thread on :368).** Trimmed in `20f81ee` to the contract; history dropped.

## Declined

**Follow-up tickets for AC2 and AC4.** You are right that they do not exist — I searched DOPE and RTOP too and found nothing. Declined by the repo owner: the deferrals are recorded on DOPE-614 and the PDU-segmentation work is being handled directly rather than tracked as separate issues.

**CRA assessment.** Not being produced for this change. Worth recording what I found, since it is broader than you raised it: SEC-02 says the assessment is "mandatory for every demand" and that the section 8 signatures "gate the merge to `development`" — there is no bug-fix exemption in the template, so the one you assumed exists does not. Two genuine ambiguities: the trigger list says "parsing of externally supplied **files**", and a debug PDU is not a file; and the process taxonomy (Epic / Demand / Subtask) never says where a Bug sits. Also relevant — DOPE-613 merged as #232 under the same reading. Both are the owner's call and both have been made.

**"No requirements document exists" line.** Declined, and I could not find the rule: it is not in *Software Development Process: Jira & GitHub*, not in the SEC-02 template, and not in either repo's PR template — strucpp's is one line, the editor's is a DOD checklist with no such field. If it is written down somewhere I did not search, point me at it and I will follow it next time.

**Assignee.** DOPE-614 assigned.

## Unchanged

The fix itself is untouched by any of this — `20f81ee` is comment-only. Full suite re-run after it: 2370 passed / 95 files, 7 skipped.

JoaoGSP (approved):
LGTM

PR #234, opened 2026-09-08
fix(codegen): use CODESYS units for temporal/integer conversions (DOPE-618)

thiagoralves:
Fixes [DOPE-618](https://autonomylogic.atlassian.net/browse/DOPE-618). Reported on the forum: [Current Date & Time Bug](https://edge.autonomylogic.com/forum/thread/current-date-time-bug) — `HOUR_OF_DT`, `MINUTE_OF_DT` and `SECOND_OF_DT` returning nonsense.

## The bug

For `DT#2026-09-02-15:36:55.360` the reporter got **11 / 23 / 44** instead of 15 / 36 / 55.

`wrapTemporalArgForNumericConversion` scaled every temporal source to milliseconds and left DATE as a raw day count. Right for TIME and TOD; wrong for DT and DATE.

**The library was correct and we were not.** OSCAT is written for CODESYS, which holds DATE, DT and TOD in a 32-bit DWORD with a 1970-01-01 epoch — DATE and DT at **seconds**, TOD at **milliseconds** — TIME in milliseconds, and the 64-bit `L` variants in nanoseconds. Its own examples pin it, and each is now a test:

| CODESYS example | value | unit |
|---|---|---|
| `DT_TO_DINT(DT#2019-9-1-12:0:0.0)` | 1567339200 | seconds |
| `DATE_TO_DINT(D#1970-1-2)` | 86400 | seconds, **not** 1 day |
| `TOD_TO_DINT(TOD#12:0:0)` | 43200000 | milliseconds |

One trap for reviewers: that page's *prose* for DATE says "days since 1970-01-01" while its own *example* returns 86400. The example agrees with the datatype page ("DATE ... handled internally as a DWORD ... seconds resolution"), so DATE is seconds and the sentence is wrong.

So OSCAT divides DT and DATE by 86400 expecting seconds, which broke **20 of its functions**: the DT family came back *aliased* (ms-since-epoch overflows a DWORD every ~49.7 days, so the answer was not even a clean factor of 1000 out) and the DATE family returned constants — `DAY_OF_WEEK` gave 4 for every date, `YEAR_OF_DATE` gave 1970 for every date.

## The fix

Both directions move together. OSCAT converts out and back inside a single expression — DCF77 computes `DWORD_TO_DT(DT_TO_DWORD(mez) - 7200)` — so fixing only the forward path would have left those callers worse off than before.

The reverse scaling used to live in the runtime's `TO_TIME(integer)` template, which **cannot** tell a millisecond count from a TIME value: every temporal type is `IECVar<int64_t>` there, so `TO_TIME(ms)` and `TO_TIME(aTimeValue)` are the same call. It now happens in codegen, where the IEC type still exists, and the runtime `TO_*` are pure casts.

The units live in one table, `TEMPORAL_CONVERSION_UNITS`, rather than a chain of conditionals, because the two directions have to agree on every row.

## Two things reviewers should know

**The `L` variants are unreachable.** The table carries LTIME / LTOD / LDT / LDATE rows and codegen has always branched on those names, but the frontend does not accept them as types — `VAR v : LDT;` is "Undefined type 'LDT'" — and they appear nowhere in `ELEMENTARY_TYPE_NAMES` or the type registry. Those branches were already dead before this change. The rows are kept because they are the right answer per CODESYS, and a test pins their absence and is written to **fail** when the types are added, since the rows become reachable then and need coverage that cannot be written today. In practice this PR fixes DT and DATE.

**Two existing tests asserted the old contract**, not a regression: `DT_TO_LINT returns milliseconds since epoch` and `DATE_TO_UINT returns days since 1970-01-01 (identity, no scaling)`. Both updated. The DATE one had been declared `UINT`, so it was quietly asserting a 16-bit truncation of the value rather than the value; it is now `UDINT`.

## Testing

`tests/integration/dope618-codesys-temporal-units-cpp.test.ts` runs the compiled program rather than asserting on emitted strings — the defect was arithmetic, and a codegen-substring test would not have caught the 32-bit aliasing that made the output look unrelated to the input.

**5 of the 7 fail on `development`.** The 2 that pass regardless are the TOD and TIME controls, which were already correct — kept so a future regression is attributed to the DT/DATE rows rather than to the scaler disappearing.

The DATE test uses five dates including a leap day and a year boundary, specifically so the old constant-output bug cannot pass it. `DAY_OF_WEEK` is asserted as ISO 8601 (Monday = 1), per OSCAT's own documentation.

Full suite **2379 passed / 96 files**. Typecheck, lint and Prettier clean.

## Verified on a real runtime

`ghcr.io/autonomy-logic/openplc-runtime:latest` (v4.2.2) in Docker, project compiled and uploaded with `openplc-cli`, compiled on the device, PLC started, values read back over the debug WebSocket. Same project and container on both sides — "before" is `development`'s strucpp:

| variable | before | after | expected |
|---|---|---|---|
| `DT_TO_DINT` | **−323863040** | 1567339200 | 1567339200 |
| `DATE_TO_DINT` | **1** | 86400 | 86400 |
| `TOD_TO_DINT` | 43200000 | 43200000 | unchanged |
| `HOUR_OF_DT` | **11** | 15 | 15 |
| `MINUTE_OF_DT` | **17** | 36 | 36 |
| `SECOND_OF_DT` | **44** | 55 | 55 |
| `YEAR_OF_DATE` | **1970** | 2026 | 2026 |
| `DAY_OF_WEEK` | **4** | 3 | 3 |
| `DAY_OF_DATE` | **0** | 20698 | 20698 |
| `HOUR` / `MINUTE` (TOD) | 15 / 36 | 15 / 36 | unchanged |
| DT round trip exact | **FALSE** | TRUE | TRUE |
| `−7200 s` → hour | **0** | 13 | 13 |

That `−323863040` is the bug in one number: milliseconds-since-epoch truncated into a signed 32-bit DINT. And `MINUTE_OF_DT` reading 17 here against the reporter's 23 — same program, only the literal's `.360` differs — is the aliasing rather than a scaling error.

## Note on shipping

Reaches users only after a strucpp release plus a `binary-versions.json` bump in openplc-editor and openplc-web — the same release DOPE-613 and DOPE-614 are waiting on. `DT_TO_*` and `DATE_TO_*` losing sub-second resolution is correct CODESYS behaviour and should be called out in the release notes; `DT_TO_MS` remains available for anyone who wants milliseconds.

[DOPE-618]: https://autonomylogic.atlassian.net/browse/DOPE-618?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/backend/codegen.ts
```diff
@@ -201,6 +201,50 @@ export interface CodeGenOptions {
  * the temporal→ms scaling: STRING / WSTRING and temporal targets need
  * different handling and stay out of this set.
  */
+/**
```
Gustavohsdp, line 197: **Problem** — the new doc block was inserted between the existing JSDoc and the `const` it belonged to. So `NUMERIC_OR_BIT_CONVERSION_TARGETS` (line 244) is now undocumented, and this interface carries the Set's documentation instead. The orphaned text is stale on top of that: it says "Used by `wrapTemporalArgForNumericConversion` to gate the temporal→ms scaling", but the Set now also gates `wrapNumericArgForTemporalConversion`, and the units it gates are no longer only milliseconds.

**Why fix it** — this file is the single source of truth for a units contract that just cost us a user-visible bug, and the comments are load-bearing: they are what stops the next person from "simplifying" the table back to milliseconds. Right now the two things a reader most needs to understand are mislabelled — the Set that decides *whether* to scale looks undocumented, and the table that decides *how much* is introduced by a paragraph about something else. A reader who trusts the surviving sentence will also conclude the reverse direction does not go through this gate, which is exactly the assumption that made DOPE-618 possible. Moving the block back onto the `const` and refreshing the sentence costs nothing and keeps the comment honest.

src/backend/codegen.ts
```diff
@@ -4527,7 +4600,7 @@ export class CodeGenerator {
         // `TO_UINT(time_var)` lowers to a `static_cast<uint16_t>(raw_ns)`
         // and the user sees the low 16 bits of the nanosecond count
         // instead of the milliseconds they asked for.
-        return this.wrapTemporalArgForNumericConversion(
+        return this.scaleConversionArg(
```
Gustavohsdp, line 4631: **Problem** — the comment immediately above this call (line 4599) still instructs the reader to "wrap the argument with `TIME_TO_MS` / `TOD_TO_MS` / `DT_TO_MS`". `DT_TO_MS` is no longer emitted on this path — it is `DT_TO_SECONDS` now — and the call no longer does only temporal→numeric: `scaleConversionArg` resolves both directions.

**Why fix it** — `DT_TO_MS` is precisely the helper whose use here *was* the bug, and it still exists in the runtime (`iec_dt.hpp:113`), so the comment does not merely go stale, it points at the wrong function by name and reads as if that were the intended behaviour. Someone debugging a future conversion problem will grep for `DT_TO_MS`, find it alive, and reasonably conclude codegen still emits it. The comment also hides half of what the call now does, so the reverse path stays undocumented at the only place a reader is looking at it.

src/backend/codegen.ts
```diff
@@ -4254,73 +4298,102 @@ export class CodeGenerator {
...
+    argExpr: string,
+    fromTypeUpper: string,
+    toTypeUpper: string,
+  ): string {
+    const scaled = this.wrapTemporalArgForNumericConversion(
+      argExpr,
+      fromTypeUpper,
+      toTypeUpper,
+    );
+    if (scaled !== argExpr) return scaled;
```
Gustavohsdp: **Problem** — this decides "did the first direction apply?" by comparing the returned string against the input. It works today, but it infers control flow from a value instead of asking the question directly. The two guards are already mutually exclusive: `wrapTemporalArgForNumericConversion` requires a numeric *target*, `wrapNumericArgForTemporalConversion` a numeric *source*, and both return the argument untouched when the type is not in `TEMPORAL_CONVERSION_UNITS`. So the comparison is not actually load-bearing — either call could simply be applied in sequence.

**Why fix it** — the failure mode is silent and it is the one this design was built to prevent. A row whose `toUnit` is a legitimate no-op (the `LTIME` / `LTOD` / `LDT` rows are exactly that shape today) returns the argument unchanged, so `scaled !== argExpr` is false and the value falls through to the *reverse* scaler. It is harmless right now only because those types are unreachable from the frontend and the second guard rejects them anyway — two accidents, not an invariant. The day the `L` types land, the test added in this PR will correctly fail for the missing coverage, but this line will quietly start routing a forward conversion into the backward one, which is a wrong number rather than a compile error. Applying both unconditionally, or branching on which side is temporal, removes the coupling to a coincidence.

src/runtime/include/iec_date.hpp
```diff
@@ -61,6 +61,30 @@ inline int64_t DATE_TO_DAYS(IEC_DATE d) noexcept {
...
+// between our day count and the two units the conversions have to produce.
+inline constexpr int64_t DATE_SECONDS_PER_DAY = 86400LL;
+inline constexpr int64_t DATE_NS_PER_DAY = DATE_SECONDS_PER_DAY * 1000000000LL;
+
+inline int64_t DATE_TO_SECONDS(IEC_DATE d) noexcept {
+    return iec_unwrap(d) * DATE_SECONDS_PER_DAY;
+}
+
+inline IEC_DATE DATE_FROM_SECONDS(int64_t s) noexcept {
+    return IEC_DATE(static_cast<DATE_t>(s / DATE_SECONDS_PER_DAY));
```
Gustavohsdp: **Problem** — `DATE_FROM_SECONDS` and `DATE_FROM_NS` (line 85) truncate toward zero, so a negative input rounds the wrong way: `DATE_FROM_SECONDS(-1)` yields day 0 (1970-01-01) where the floor is day −1 (1969-12-31). `DATE_t` is `int64_t` (`iec_types.hpp:131`), so the value is representable and nothing rejects it.

**Why fix it** — `DATE_OF_DT` in `iec_dt.hpp:94-100` already handles exactly this borrow explicitly, with a comment saying so, so the runtime now answers the same question two different ways depending on which helper you reach. Pre-1970 DATE is outside CODESYS's documented domain, which is why this is a nit and not a blocker — but these helpers are fed by codegen from arbitrary user integer expressions (`DINT_TO_DATE(x)`), not only from literals, so the input is not actually bounded by the domain. Matching `DATE_OF_DT`'s floor division, or stating in a comment that negative input is out of contract, keeps the two from drifting further apart.

src/runtime/include/iec_date.hpp
```diff
@@ -61,6 +61,30 @@ inline int64_t DATE_TO_DAYS(IEC_DATE d) noexcept {
     return iec_unwrap(d);
 }
 
+// DATE is stored as whole days, but CODESYS holds DATE in the same memory
+// format as DT — seconds since 1970-01-01, truncated to a day boundary — and
+// its DATE_TO_* conversions yield that: `DATE_TO_DINT(D#1970-01-02)` is 86400,
+// not 1. LDATE is the 64-bit form and yields nanoseconds. These four convert
+// between our day count and the two units the conversions have to produce.
+inline constexpr int64_t DATE_SECONDS_PER_DAY = 86400LL;
+inline constexpr int64_t DATE_NS_PER_DAY = DATE_SECONDS_PER_DAY * 1000000000LL;
```
Gustavohsdp: **Problem** — this is the third copy of the same constant in the runtime headers: `DT_NS_PER_DAY` (`iec_dt.hpp:49`) and `TOD_NS_PER_DAY` (`iec_tod.hpp:46`) are both `24 * 60 * 60 * 1e9`.

**Why fix it** — no objection to it landing as is; this is a pre-existing pattern rather than something the PR introduced, and the constant is not going to change. Raising it only because this PR is the first time all three are used for *unit conversion between* the types rather than internally to one of them, which is when a divergence would start producing wrong values instead of just duplicated text. One shared definition while this area is open is cheap; if you would rather not touch three headers in a bug fix, that is a perfectly good answer and I would leave it.

Gustavohsdp (commented):
Reviewed by running the branch, not only by reading the diff. The fix is correct and the evidence holds up.

## Verified locally (`e059c5b`)

- `npm run typecheck` clean; lint 0 errors (506 pre-existing warnings); Prettier clean on the changed `.ts` files.
- Full suite: **96 files / 2379 passed / 7 skipped** — matches the PR body exactly.
- **Negative control**: applied the new test file on top of `development` → **5 failed / 2 passed**, reproducing `11 23 44`, `-323863040`, `YEAR_OF_DATE = 1970` and `DAY_OF_WEEK = 4` for every date. The central claim of this PR stands.
- OSCAT premise corroborated independently: the string table inside `oscat_basic_335.library` carries `DT_TO_DWORD` (14x), `DATE_TO_DWORD` (18x), `DWORD_TO_DT` (8x) and the constants `86400`, `946728000`, `31557600`, `3600000` — including the reverse direction, which is what justifies moving both paths together.
- Probed 11 expression shapes (array element, struct field, FB output, user function, bit access, dereference, nested conversion, `SHL`, bare literal, ...) looking for a gap left by removing the `* 1e6` from the runtime `TO_TIME`: **all** of them emit `TIME_FROM_MS`. I could not construct a reachable path where the scaling is lost.

**Not verified**: the "Verified on a real runtime" table (Docker + `openplc-cli` + debug WebSocket) — I did not run the runtime.

## Process — owed, but no reason to hold the code

- The PR body links the Jira issue but does not state that this is a bug fix with no Requirements Gathering document and no Cybersecurity Risk Assessment. **Why it matters:** the process allows the exemption, not the silence. A later reader — or an audit — cannot tell "exempt" from "skipped" without that line, and the safe assumption they will make is the wrong one.
- DOPE-618 is still in **Backlog with no assignee** while this PR is open. **Why it matters:** the board stops describing reality, so anyone picking up work sees this bug as untouched.
- AC 8 (the `L` variants) is answered here as "unreachable", with a test that fails the day the types are accepted. That is a well-argued deviation — but the ticket still lists the criterion as owed. **Why it matters:** the task is the copy most people actually read; leaving the criterion open there sends the next person hunting for coverage that was deliberately not written.
- AC 6 (release notes) cannot be satisfied inside this repo — there is no CHANGELOG or release-notes file. **Why it matters:** `DT_TO_*` and `DATE_TO_*` losing sub-second resolution is a public behaviour change. If it is not on the release checklist now, it ships silently and the next forum report is ours.

Nothing below blocks the merge.

src/backend/codegen.ts
thiagoralves: Confirmed and fixed in `8c99662`. My insertion point was simply wrong — I put the new block between the Set and its own JSDoc, so the Set read as undocumented and my table looked documented by a paragraph about the Set.

Moved the block back onto `NUMERIC_OR_BIT_CONVERSION_TARGETS` and rewrote it, since the old text was stale in both the ways you name: it described only `wrapTemporalArgForNumericConversion` and only ms scaling. It now says the Set gates BOTH directions and how — a temporal source scales only when the target is in the Set, a numeric source scales into a temporal target only when the source is — and records that the Set and the table are disjoint, which is the property that makes the two directions mutually exclusive.

src/backend/codegen.ts
thiagoralves: Fixed in `8c99662`. Verified your point exactly: that comment was the only remaining `DT_TO_MS` anywhere in codegen, and the function is still alive at `iec_dt.hpp:113` — so the grep you describe lands on a live function and the wrong conclusion.

Rewritten to name neither helper. It now says the scaling applies in whichever direction fits, points at `TEMPORAL_CONVERSION_UNITS` for the units and why they are CODESYS's, and keeps the part that actually belongs at a call site: why this cannot live in the runtime, since every temporal type is the same C++ type there.

src/backend/codegen.ts
thiagoralves: Refactored in `8c99662` — it now branches on which side is temporal rather than comparing the returned string against its input.

The style point is right and the fix is worth having. But **the failure mode you describe does not hold**, and I checked rather than assumed: I extracted both sets from the source and their **intersection is empty**. `NUMERIC_OR_BIT_CONVERSION_TARGETS` is the 15 numeric/bit types; `TEMPORAL_CONVERSION_UNITS` is the 13 temporal spellings. So for a forward conversion the reverse guard requires the SOURCE to be numeric, and a temporal source never is — every one of the five no-op rows (`LTIME`, `LTIME_OF_DAY`, `LTOD`, `LDT`, `LDATE_AND_TIME`) is temporal-only.

That is not "two accidents" — it is structural: a type cannot be both a temporal type and a numeric target, and adding the L types to the frontend would add them as temporal types, not numeric ones. So the fall-through returned the argument unchanged, which is the correct answer for a pass-through row, and would keep doing so after the L types land.

Taken anyway, on your readability argument: inferring "did this direction apply?" from a value comparison is fragile to read even when it is correct, and the explicit form does not depend on two separate tables continuing to agree. The commit message records that the predicted wrong-number case was checked and did not exist, so nobody re-derives it later.

src/runtime/include/iec_date.hpp
thiagoralves: Confirmed and fixed in `8c99662`. Measured before touching it:

```
DATE_FROM_SECONDS(-1)     = 0    <- 1970-01-01, floor is -1
DATE_FROM_SECONDS(-86400) = -1
DATE_FROM_NS(-1)          = 0
DATE_OF_DT(DT = -1ns)     = -1   <- floors, with the explicit borrow
```

So exactly the split you describe: the same question answered two ways depending on which helper you reach.

Both helpers now floor, through a shared `iec_floor_div` that carries the reasoning — including your point that the domain argument does not bound the input, because these take arbitrary user expressions via `DINT_TO_DATE(x)` rather than only literals.

Two tests, and I checked both discriminate. One asserts the value (`-86400 -86400`; truncation gives `0 -86400`). The other asserts that `DINT_TO_DATE` and `DATE_OF_DT` **agree with each other** rather than with a hard-coded number, so the two cannot drift apart again — that one reports `DISAGREE` with truncation restored.

src/runtime/include/iec_date.hpp
thiagoralves: Taken, in `8c99662` — your reasoning for raising it is the reason to do it now rather than leave it. This change is the first to convert BETWEEN the three types, so a divergence would start producing wrong values instead of duplicated text, and that is precisely when the duplication stops being harmless.

`IEC_SECONDS_PER_DAY` and `IEC_NS_PER_DAY` now live in `iec_types.hpp`, which all three headers already include, so it cost no new dependency. `DT_NS_PER_DAY`, `TOD_NS_PER_DAY`, `DATE_NS_PER_DAY` and `DATE_SECONDS_PER_DAY` are kept as aliases of it, so **no call site changed** — which keeps the blast radius of a bug fix touching three headers down to four one-line declarations.

`static_assert`s in the test driver pin all four to the shared constant, so a future divergence fails at compile time rather than becoming a wrong date.

Gustavohsdp (approved):

PR #237, opened 2026-09-16
fix(lsp): stop emitting overlapping semantic tokens for arrays (RTOP-287)

JoaoGSP:
Fixes **RTOP-287**, split out of DOPE-544 where the OpenPLC editors were logging `Overlapping semantic tokens detected` after any programmatic ST body rewrite. The client was cleared first: the editors only window and rebase the token stream, never inspect a column, so the overlap originates here.

A client that receives overlapping ranges **drops tokens and warns**, so this was never purely cosmetic — highlighting silently degraded on any line that declares or indexes an array.

## Three changes

**1. `getChildren` visited every subscript index twice** (`src/ast-utils.ts`)

A `VariableExpression` pushed both `ve.subscripts` and the indices carried by `ve.accessChain`. Those are separate clones of the same indices, not shared references, so `walkAST` reached each one twice and the emitters pushed a duplicate token.

This is the wider of the three: it means **references and rename were counting subscript index variables twice too**. Both of their test suites still pass, and they get more accurate as a side effect.

**2. `emitTypeReference` painted over the variable name** (`vscode-extension/server/src/semantic-tokens.ts`)

An inline array's `TypeReference` is synthesised — `__INLINE_ARRAY_INT`, `__VLA_…` — and carries the *whole declaration's* span. `emitTypeReference` pushed `tr.name.length` characters from `tr.sourceSpan.startCol`, so it painted 18 characters from the declaration's start column and swallowed the variable name:

```
line 3 col 3  "Arr"(variable) [2,5)  vs  "Arr : ARRAY [1..10"(type) [2,20)
```

The highlighted width was an accident of how long the synthetic name happened to be. Synthetic names are now skipped — no token can be correct for a name that does not appear in the source.

I deliberately did **not** change the name or the span in `ast-builder.ts`. `__INLINE_ARRAY_<T>` is load-bearing for codegen (`IEC___INLINE_ARRAY_<T>` in `project-model.ts` and `codegen.ts`), so the narrow fix belongs in the emitter.

**3. Defensive clip in `deltaEncode`**

After the sort, fully covered tokens are dropped and partial overlaps clipped. This makes the class unreachable from the client whatever the emitters do later.

## Tests

New `vscode-extension/tests/unit/semantic-tokens-overlap.test.ts` asserts the client's own invariant — within a line, no token may start before the previous one ends — over 8 constructs plus both bundled fixtures, and adds two behavioural assertions: an array declaration line carries only the variable name as a `variable`, and a subscript index is emitted exactly once.

Before this branch, on `development` @ `302c7b6`:

| Case | Overlaps |
| --- | --- |
| array declaration | 1 |
| array + subscript | 2 |
| 2-D array | 1 |
| array of user-defined type | 1 |
| literal subscript | 2 |
| subscript on the right-hand side | 2 |
| no array (control) | 0 |
| `complex-project.st` / `simple-program.st` | 0 |

Both fixtures were already clean, which is why this went unnoticed — neither contains an inline array declaration.

## Validation

| Suite | Result |
| --- | --- |
| new overlap test | 12 / 12 |
| `vscode-extension` | 294 / 294 |
| core `npm test` | 2398 passed, 7 skipped, 96 files |

The core suite matters here because change 1 lands in the shared package.

## After merge

Needs a strucpp release plus a `binary-versions.json` bump in openplc-web and openplc-editor before the editors see it.

https://claude.ai/code/session_01P2Wpvri7jZKGPUqgWrpAkE


vscode-extension/server/src/semantic-tokens.ts
```diff
@@ -290,6 +290,8 @@ function emitTypeReference(
   if (!tr.sourceSpan) return;
   // Skip REF_TO/REFERENCE_TO keyword-prefixed references (the type name position is offset)
   if (tr.isReference) return;
+  // Synthesised inline-array names (`__INLINE_ARRAY_INT`, `__VLA_…`) have no source text to cover.
+  if (tr.name.startsWith("__")) return;
```
dcoutinho1328: [P2] Restrict the filter to synthetic array types

The `startsWith("__")` guard also suppresses real types, including the supported elementary type `__XWORD`. For a declaration such as `addr : __XWORD;`, the previous implementation emits a `type` token with the `defaultLibrary` modifier; this change removes that token entirely. User-defined types such as `__Count` lose their highlighting too. Please restrict the filter to the synthetic `__INLINE_ARRAY_` and `__VLA_` prefixes, or use explicit synthetic-type metadata, and add a regression test preserving `__XWORD` highlighting.

JoaoGSP: Good catch — taken in e4d9b23. You are right that it was a regression, and it was worse than the elementary type alone.

Verified before fixing: `__XWORD` is the **only** underscore-prefixed key among the 24 in `ELEMENTARY_TYPES` (aliased in `src/semantic/type-utils.ts:59`, and load-bearing in `type-codegen.ts` and `debug-table-gen.ts`). Probing the built extension confirmed the effect — `addr : __XWORD;` emitted only `"addr"(variable)`, with the `type`/`defaultLibrary` token gone. A user type opening with underscores lost its token the same way: `c : __Count;` emitted only `"c"(variable)`.

The guard now matches the two prefixes the AST builder actually mints (`ast-builder.ts:1606-1607`), held in a named constant so the reason they are special is visible where they are listed rather than only at the call site:

```ts
const SYNTHETIC_TYPE_PREFIXES = ["__INLINE_ARRAY_", "__VLA_"];
```

Two regression tests added to `semantic-tokens-overlap.test.ts`, both of which fail against the previous guard:

* `addr : __XWORD;` → `["addr"(variable), "__XWORD"(type)]`
* `c : __Count;` → `["c"(variable), "__Count"(type)]`

I did not go to explicit synthetic-type metadata on the AST node, which was your other suggestion. It is the more robust answer and I would take it if these prefixes multiply, but it means touching the core package that codegen depends on, and this PR is otherwise confined to the extension — so the prefix list plus the tests felt like the right trade here. Happy to switch if you would rather have the metadata.

Full suites after the change: core 2398 passed / 7 skipped (96 files), extension 296 passed (21 files), `tsc --noEmit` clean.

PR #238, opened 2026-09-17
feat(debug): address a leaf's value in place instead of copying it out

thiagoralves:
## What

Adds a sixth op to `TypeOps` in `debug_dispatch.hpp`:

```cpp
const void* (*ptr)(const void*, uint16_t* len);
```

and `handle_ptr(arr, elem, out_len)` over it. It returns a pointer to the leaf's live storage plus the value's **current** length — not its padded wire width — so a caller that can serve a value without owning it does not pay an allocation and a copy per read.

Scalars go through `IECVar<T>::read_ptr()`, a new accessor added alongside. **`read_ptr()` and not `raw_ptr()`**: a located variable is written by the PLC program straight into `value_`, so `raw_ptr()` would leak the program's value past an active force. STRING/WSTRING go through `IECStringVar::c_str()` / `length()`, with a `static_assert` that the target is little-endian since a WSTRING is handed over as UTF-16LE bytes.

## Why

`read_node` in the baremetal OPC-UA server read into a `uint8_t buf[8]` and `UA_Variant_setScalarCopy`'d out of it, allocating from a ~19 KB arena on every value of every read — and it could not hold a string at all, so STRING and WSTRING were unservable. With this op a read is allocation-free and a 253-byte WSTRING fits.

## The safety contract, and why this is NOT exported to Linux

`handle_ptr()` is deliberately **not** part of the `strucpp_debug_*` C exports. The pointer is valid only until the caller yields to the scan. That holds on a cooperative single-threaded super-loop (the baremetal runtime services OPC-UA in the tail of the cycle, so the program cannot move the value underneath it) and is false for the Runtime v4 `.so`, where the scan runs in its own thread and a plugin reading through the pointer would race it. Linux callers keep using `strucpp_debug_read()`.

## Release ordering — please read

`openplc-editor` **already depends on this** (`arduino_runtime_glue.cpp` calls `strucpp::debug::handle_ptr`), and its `binary-versions.json` still pins `v0.6.7`, which does not have it. So until this merges and a release is cut:

1. merge this
2. cut a release
3. bump `binary-versions.json` in openplc-editor / openplc-web

A clean editor checkout will fail to compile firmware in between. Validated locally by packing this branch (`npm pack`) and installing it the way `downloadStrucpp()` does (`npm install <tgz> --no-save`), then building and flashing a Siemens LOGO! 8.2.

## Testing

Full strucpp suite green (96 files, 2399 passed). End-to-end on a LOGO! 8.2 through the OPC-UA server: STRING and WSTRING read and written, empty strings round-tripped, both strings in a single batched Read returning distinct correct values, and every write cross-checked over the Modbus debug channel.

Refs openplc-editor#1091, Autonomy-Logic/openplc-packages#48, DOPE-645

https://claude.ai/code/session_01NegApcNt2SzRk3FDAs5T9p


src/runtime/include/debug_dispatch.hpp
```diff
@@ -236,6 +262,24 @@ inline void unforce_wstring(void* p) noexcept {
...
 }
 
+// The code-unit buffer as it sits in memory, so `len` is BYTES (2 per code
+// unit), not characters.
+//
+// read_wstring() splits each code unit explicitly to keep the WIRE format
+// little-endian on any host. A pointer cannot do that — what the caller gets is
+// host order — so this op is only correct on a little-endian target, and says
+// so rather than silently serving byte-swapped text somewhere else.
+static_assert(__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__,
```
Gustavohsdp: **This breaks the build on big-endian targets, not just `ptr_wstring`.**

The assert is at namespace scope and unconditional, so it fires when the header is *included*, not when the op is used. Any TU that pulls in `debug_dispatch.hpp` on a big-endian target now fails to compile — including AVR/Modbus firmware that only ever calls `handle_read`/`handle_write`. And `read_wstring()` right above was deliberately written with an explicit byte split so the wire form stays LE "on any host". The comment tells a big-endian caller to "copy through `read_wstring()` instead" — after this change, that caller cannot build at all.

Verified by compiling this branch's header with a simulated big-endian `__BYTE_ORDER__`, with nothing referencing `ptr_wstring`:

```
inc/debug_dispatch.hpp:272:15: error: static assertion failed due to requirement '4321 == 1234':
  ptr_wstring hands out UTF-16 code units in host order; a big-endian target must
  copy through read_wstring() instead.
```

Suggested shape — keep the op honest without taking the header down with it:

```cpp
#if defined(__BYTE_ORDER__) && __BYTE_ORDER__ != __ORDER_LITTLE_ENDIAN__
// A pointer cannot byte-split the way read_wstring() does, so on a big-endian
// target there is no correct answer to give: refuse, and let the caller fall
// back to handle_read() / strucpp_debug_read().
inline const void* ptr_wstring(const void*, uint16_t* len) noexcept {
    *len = 0;
    return nullptr;
}
#else
inline const void* ptr_wstring(const void* p, uint16_t* len) noexcept { /* as written */ }
#endif
```

Secondary: `__BYTE_ORDER__` is a GCC/Clang predefined macro, not defined by MSVC, where the expression is a hard error regardless of endianness. The `defined(...)` guard covers that too.

src/runtime/include/debug_dispatch.hpp
```diff
@@ -187,6 +204,15 @@ inline void unforce_string(void* p) noexcept {
     static_cast<IECStringVar<254>*>(p)->unforce();
 }
 
+// The payload only — NOT the [len][payload] wire form read_string() builds.
+// `len` is the character count, and c_str() already resolves the force.
+inline const void* ptr_string(const void* p, uint16_t* len) noexcept {
```
Gustavohsdp, line 224: **`STRING(n)` leaves come back empty through this op — and through `read_string` too. Pre-existing, but this PR extends it to OPC-UA.**

`ptr_string` casts to `IECStringVar<254>*`, but codegen emits the declared size — `src/backend/codegen.ts:565` returns `IECStringVar<${maxLength}>` — and the debug table tags the leaf `STRING` regardless (`src/backend/debug-table-gen.ts:60`). `IECStringVar<N>::length_` sits at a different offset for every `N`, so the cast reads it out of bounds.

Measured on this branch, program `VAR s20 : STRING(20) := 'hello'; sdef : STRING := 'world'; END_VAR`:

```
S20  ptr_len=0 payload=''
S20  read_wire_len=0 payload=''
SDEF ptr_len=5 payload='world'
```

`sizeof(IECStringVar<20>)` is 50 and `length_` is read at offset ~255. It landed on 0 here, but the value is undefined. In `opcua_nodes.cpp:read_node` that `len` goes straight into `hdr->length` under `UA_VARIANT_DATA_NODELETE`, so a nonzero garbage length <= 126 would have open62541 encode bytes past the end of a 21-byte buffer onto the wire.

Not introduced here — `read_string`, `write_string` and `force_string` carry the same cast and the Modbus debug path already lives with it — so I would not block the merge on fixing it in this PR. It does deserve its own issue, and the description's claim that STRING/WSTRING are now servable holds only for strings declared without a length. The fix is upstream of the dispatch: either have codegen always emit `IECStringVar<254>` for debug-registered strings, or carry the declared `MaxLen` in the `Entry` and dispatch on it.

tests/integration/debug-table-cpp.test.ts
```diff
@@ -467,6 +467,116 @@ int main() {
...
+  // straight to a protocol stack (OPC-UA gives open62541 a UA_Variant marked
+  // NODELETE). The properties that matter are that it addresses the SAME value
+  // handle_read() would produce — including through a force — and that a STRING
+  // comes back as its payload rather than the [len][payload] wire form.
+  //
+  // The force case is the one that motivated read_ptr(): force() writes through
+  // to value_, so raw_ptr() agrees the instant a force is applied — but a
+  // located variable is driven by the program straight into value_ every scan,
+  // and then only read_ptr() still resolves to the forced value.
+  it("addresses the same value handle_read copies, force included", () => {
```
Gustavohsdp, line 481: **`ptr_wstring` has no test.**

This covers `ptr_impl<DINT>` and `ptr_string` well, including the force-beats-a-direct-program-write case that motivates `read_ptr()`. But `ptr_wstring` — the one op with non-trivial arithmetic (`*len = capped * 2`, bytes rather than code units) and the one carrying an endianness precondition — is exercised only on the bench, and the LOGO! 8.2 run is not reproducible in CI.

Cheap additions to this same test:

- a `WSTRING` leaf: `len` is `2 x character count`, and the code units match what `read_wstring()` put on the wire once un-split;
- an empty `STRING` and an empty `WSTRING` (`len == 0`, non-null pointer) — `read_node` explicitly treats zero length as a valid value, so it is a contract, not an edge case;
- a forced `STRING` through `handle_ptr`, mirroring the scalar force assertion. `ptr_string` resolves force through `c_str()`, and nothing here proves it.

src/runtime/include/debug_dispatch.hpp
```diff
@@ -459,6 +507,25 @@ inline uint16_t handle_size(uint8_t arr, uint16_t elem) noexcept {
...
 }
 
+/** Address a leaf's value in place; `out_len` receives its length in BYTES.
+ *  Returns nullptr (and sets *out_len = 0) for an out-of-range leaf.
+ *
+ *  The copy-free counterpart to handle_read(). The pointer is into live PLC
+ *  storage, so it is valid only until the variable is next written — safe for a
+ *  caller that runs inside the scan's own thread of control, wrong for one that
+ *  can be preempted by it. */
+inline const void* handle_ptr(uint8_t arr, uint16_t elem, uint16_t* out_len) noexcept {
```
Gustavohsdp, line 603: **The lifetime contract is missing `unforce()`.**

"valid only until the variable is next written" is not the whole rule. `read_ptr()` returns `&forced_value_` while forced and `&value_` otherwise (`iec_var.hpp:221`), and `IECStringVar::c_str()` does the same — so **changing the force state also invalidates a retained pointer**, silently: after `unforce()`, a pointer taken while forced keeps reporting the stale forced value even though the variable is no longer forced.

The current caller re-reads per Read service call, so it cannot hit this. The comment is what the next caller will trust. Suggest "valid only until the variable is next written **or its force state changes**", here and at `debug_dispatch.hpp:132` and `iec_var.hpp:207`.

src/runtime/include/debug_dispatch.hpp
```diff
@@ -245,6 +289,10 @@ struct TypeOps {
     void (*unforce)(void*);
     void (*read)   (const void*, uint8_t*);
     void (*write)  (void*, const uint8_t*);
+    /** Address the value in place instead of copying it out, and report its
+     *  length in BYTES. Same value `read` would produce, same force
+     *  resolution — see ptr_impl(). */
+    const void* (*ptr)(const void*, uint16_t*);
```
Gustavohsdp: `handle_ptr` null-checks `out_len`, but `ptr_impl`, `ptr_string` and `ptr_wstring` all dereference `len` unconditionally. Fine while `handle_ptr` is the only caller and always passes a real address — worth stating it here, e.g. "`len` must be non-null".

src/runtime/include/debug_dispatch.hpp
```diff
@@ -253,27 +301,27 @@ struct TypeOps {
 // Kept inline so it's flash-resident with no separate .cpp required.
 // ---------------------------------------------------------------------------
 inline constexpr TypeOps type_ops[TAG__COUNT] = {
```
Gustavohsdp, line 341: **Footprint (a question, not a claim).** The `Entry` tables carry `STRUCPP_DEBUG_FLASH` (`debug_table.hpp:158`); `type_ops` does not, so on AVR it lands in SRAM rather than staying flash-resident as the comment above says. The new column adds one function pointer per row — on the order of 42 bytes of SRAM on an ATmega-class part — plus 21 `ptr_impl<T>` instantiations in flash, and every AVR firmware pays it although only the baremetal OPC-UA server calls `handle_ptr`. I could not measure it (the local avr-gcc ships no C++ headers): did the smallest supported target's footprint move?

**Formatting.** The scalar rows lost their column alignment while the STRING/WSTRING rows kept theirs, and the trailing comment padding changed (`/*STRING  */` -> `/*STRING */`). It turns a one-column change into 21 rewritten lines and makes the table harder to scan.

src/runtime/include/iec_var.hpp
```diff
@@ -204,6 +204,22 @@ class IECVar {
...
+     * the PLC program every scan, and the program writes value_ directly through
+     * the image binding, not through set(). raw_ptr() would then show the
+     * program's value while get() still reports the forced one.
+     *
+     * So this is get()'s semantics with get()'s copy removed, which is what an
+     * external reader (the debug dispatch's pointer op, and through it OPC-UA)
+     * needs to serve a value without copying it. Mirrors IECStringVar::c_str(),
+     * which resolves the force the same way.
+     */
+    const T* read_ptr() const noexcept { return forced_ ? &forced_value_ : &value_; }
```
Gustavohsdp, line 221: `docs/RUNTIME.md` lists `IECVar`'s public surface (lines 44–73) and its "Located Variables" section explains `raw_ptr()` — exactly the distinction this comment draws. `read_ptr()` should join it, otherwise the rule is written down only in a header comment.

Gustavohsdp (commented):
## Full review

**Verdict: needs changes.** The design is sound and the safety reasoning is the right one — `read_ptr()` over `raw_ptr()`, and deliberately not exporting `handle_ptr` to Runtime v4. Three things should land before merge, plus the process links.

| Severity | Where | Problem | Confidence |
|---|---|---|---|
| blocker | `debug_dispatch.hpp:272` | Unconditional namespace-scope `static_assert` breaks the build on any big-endian target, including firmware that only calls `read_wstring()` | 9/10 |
| required | `debug_dispatch.hpp:209` | `STRING(n)` leaves read back as length 0 — the ops cast to `IECStringVar<254>*` while codegen emits `IECStringVar<n>`. Pre-existing, extended to OPC-UA here | 10/10 |
| required | `debug-table-cpp.test.ts:481` | `ptr_wstring` has no test at all — the `*2` byte arithmetic and the force path are uncovered | 9/10 |
| required | PR description | No Jira issue of its own, no Requirements Gathering link, no Cybersecurity Risk Assessment link | 9/10 |
| required | `debug_dispatch.hpp:517`, `iec_var.hpp:221` | The documented lifetime contract misses `unforce()`, which also invalidates a retained pointer | 8/10 |
| nit | `docs/RUNTIME.md:44` | `read_ptr()` joins `IECVar`'s public surface; the doc is not updated | 8/10 |
| nit | `debug_dispatch.hpp:209`, `:275` | Silent truncation at `DEBUG_STRING_CAP` (126) on a transport with no 126-byte frame limit — intended, or inherited? | 7/10 |
| nit | `debug_dispatch.hpp:303` | `type_ops` is not `STRUCPP_DEBUG_FLASH`, so the new column costs SRAM on AVR for firmware that never calls `handle_ptr` | 6/10 |
| nit | `debug_dispatch.hpp:304` | Scalar rows lost their column alignment while STRING/WSTRING kept theirs — 21 lines of diff for a 1-column change | 7/10 |
| nit | `debug_dispatch.hpp:295` | `len` is dereferenced unconditionally in all three ops while `handle_ptr` null-checks `out_len` | 7/10 |

Details are inline on the lines they belong to.

### Overlap with DOPE-645 worth deciding now

DOPE-645 proposes `handle_actual_size()` in strucpp for the same padded-vs-actual length problem this op solves from the other end. Two ways to learn a string's real length will exist. Worth settling whether `handle_ptr`'s `out_len` is the answer for both, or whether they coexist on purpose — before the second one is written.

### Checks

- ✅ `handle_ptr` is genuinely absent from the `strucpp_debug_*` C exports
- ✅ Consumer contract matches openplc-editor#1091: WSTRING → `UA_TYPES_BYTESTRING`, `hdr->length = len` in bytes, `UA_VARIANT_DATA_NODELETE`, one header slot per node in a batch, zero length not treated as an error
- ✅ `read_ptr()` resolves force correctly and mirrors `IECStringVar::c_str()`
- ✅ No earlier reviews or inline comments to re-check
- ✅ `tests/integration/debug-table-cpp.test.ts` — 15/15 pass locally on this branch
- ✅ Release ordering is documented in the description, which is the right call
- ❌ CI did not run: `.github/workflows/ci.yml` triggers on `main` only and this PR targets `development`. Nothing was linted, type-checked or tested automatically. Repo config, not this PR — but it means the local run is the only evidence
- ❌ No Jira issue of its own. `DOPE-645` is a different bug (Backlog: compact framing + `handle_actual_size()`), `RTOP-285` is the editor task
- ❌ No Requirements Gathering link, no Cybersecurity Risk Assessment link, and no explicit "small change, no requirements document" line
- ➖ Acceptance criteria: neither DOPE-645 nor RTOP-285 states any

src/runtime/include/debug_dispatch.hpp
thiagoralves: **Fixed** — `afcd67c`.

Reproduced both halves first: compiling the header on a simulated big-endian target with nothing referencing `ptr_wstring` gives

```
debug_dispatch.hpp:272:15: error: static assertion failed due to requirement '4321 == 1234'
```

and with `__BYTE_ORDER__` undefined (the MSVC case) it is `error: use of undeclared identifier '__BYTE_ORDER__'` regardless of endianness. So you were right that an op nobody referenced was taking the whole translation unit down.

It is now a compile-time selection rather than an assertion. On a target that cannot be served, `ptr_wstring` returns `nullptr` with `*len = 0` — already `handle_ptr`'s contract for "no pointer available" — which routes the caller to the copying path the old assertion text told them to use anyway. Verified both ways: the big-endian probe compiles clean, the native build is unchanged.

src/runtime/include/debug_dispatch.hpp
thiagoralves: **Fixed** — `afcd67c`. Valid, and I reproduced your measurement and then a worse case.

Your numbers on this branch: `sizeof(IECStringVar<20>)` = 50, `<254>` = 518, and a `STRING(20)` holding `'hello'` comes back `len=0`. Re-running it with non-zero neighbouring memory gives **`len=126`, and 126 bytes of adjacent memory as the payload** — so "it landed on 0 here" was luck, exactly as you said.

`Entry` is `{void* ptr; uint8_t tag; uint8_t flags}` with nowhere to carry a per-leaf width, and the ops are one function per TAG, so making this work needs a real change to a structure that is deliberately 6 bytes on AVR. Rather than register a leaf the dispatch cannot address, the generator now **skips** it with a reason, and `skipped` is surfaced as a build warning (it was collected and dropped before, so a variable simply vanished from the debugger and OPC-UA with nothing to say why):

```
INSTANCE0.SIZED is not debuggable: declared STRING(20), but the debug dispatch
addresses strings as STRING(254). Declare it as a plain STRING to expose it.
```

One correction to the impact assessment, in your favour on reachability: nothing authored in the **editor** can produce a parameterised string — its type picker is a closed list from the IEC registry, which carries `STRING` with no length parameter, and the ST variable parser's type pattern excludes parentheses. So this is hand-written ST through the CLI, not a live path for editor users.

tests/integration/debug-table-cpp.test.ts
thiagoralves: **Fixed** — `afcd67c`. You were right that the one op with non-trivial arithmetic and an endianness precondition was the one with no coverage.

Added to the same test, all three of your suggestions:

- a `WSTRING` leaf: `len` is `2 x code units` (bytes, not units), and the code units match what `read_wstring()` put on the wire once its explicit LE split is undone;
- an empty `STRING` and an empty `WSTRING` — `len == 0` with a **non-null** pointer, since `read_node` treats zero length as a value, so it is a contract rather than an edge case;
- a forced `STRING` through `handle_ptr`, mirroring the scalar force assertion — `ptr_string` resolves force through `c_str()` and nothing proved it.

Confirmed they actually catch the regression: removing the `* 2` from `ptr_wstring` fails with `FAIL wstring len is bytes not units`.

src/runtime/include/debug_dispatch.hpp
thiagoralves: **Fixed** — `afcd67c`. Valid: `read_ptr()` is `forced_ ? &forced_value_ : &value_`, so `force()`/`unforce()` changes *which object* the pointer refers to, and nothing is written to invalidate it.

The contract now reads "valid only until the variable is next written **or its force state changes**", at both `handle_ptr`'s docblock and `ptr_impl`'s, and `docs/RUNTIME.md` carries the same sentence. As you said, the current caller re-reads per Read service call and cannot hit it — this is for the next caller.

src/runtime/include/debug_dispatch.hpp
thiagoralves: **Fixed** — `afcd67c`. Confirmed the asymmetry is real (`handle_ptr` guards `out_len`, the three ops dereference `len` unconditionally) and unreachable today, since `handle_ptr` is the only caller and always passes `&len`.

Documented rather than guarded, as you suggested: each of `ptr_string` and `ptr_wstring` now states "`len` must be non-null; handle_ptr always passes a real address."

src/runtime/include/debug_dispatch.hpp
thiagoralves: **Measured, and fixed** — `e38fbeb`. Answering the question you could not: yes, it moved the smallest supported target.

The Arduino AVR toolchain ships no C++ standard headers, which is what blocked you (and me, at first). The editor carries the shim at `resources/sources/avr-libstdcpp/include`, and its simulator already compiles for `arduino:avr:mega`. Same sketch, same runtime tree, a firmware calling only `handle_read`/`handle_write`:

| | SRAM | flash |
|---|---|---|
| without the `ptr` column | 383 B | 4654 B |
| with it | 425 B | 5128 B |

So **+42 bytes of SRAM and ~474 of flash on every AVR firmware**, for an op only the baremetal OPC-UA server calls — your estimate was right to the byte. And you were right about the cause: `type_ops` carries no `STRUCPP_DEBUG_FLASH`, so on a Harvard target it is `.rodata` copied into SRAM at startup; "inline constexpr" does not make it flash-resident, and the comment above it claimed otherwise.

The fix is not to move `type_ops` to PROGMEM — that routes every row read through `pgm_read_ptr` in the hot path of `handle_read`/`handle_write`/`handle_set`, which deserves its own evidence. Instead the pointer op stops being a column: `ptr_ops[]` is a table of its own, read only by `handle_ptr`, so `-ffunction-sections`/`-fdata-sections` + `--gc-sections` drop both it and the `ptr_impl<T>` instantiations when nothing calls it. Measured after:

| | SRAM | flash |
|---|---|---|
| never calls `handle_ptr` | 383 B | 4728 B |
| calls `handle_ptr` | 425 B | 5196 B |

The feature now costs what it costs, and only to firmware that uses it.

**Formatting** — also fixed in `afcd67c`: the scalar rows are realigned so the table reads as one block again.

src/runtime/include/iec_var.hpp
thiagoralves: **Fixed** — `afcd67c`. Valid: the rule lived only in a header comment.

`docs/RUNTIME.md` now lists `read_ptr()` on `IECVar`'s public surface and adds a "`raw_ptr()` vs `read_ptr()`" section next to Located Variables — a table of which points at what and which to use for binding vs reading, the located-variable case that makes them differ, and the lifetime clause including the force-state change from the thread above.

Gustavohsdp (approved):

